PR #2, opened 2021-12-29
DM-30891: stand up package with all DRP pipelines

TallJimbo:


pipelines/DECam/Merian.yaml
```diff
@@ -0,0 +1,225 @@
...
+                                                   'r': 'phot_g_mean',
+                                                   'i': 'phot_g_mean',
+                                                   'z': 'phot_g_mean',
+                                                   'Y': 'phot_g_mean',
+                                                   'N419': 'phot_g_mean',
+                                                   'N540': 'phot_g_mean',
+                                                   'N708': 'phot_g_mean',
+                                                   'N964': 'phot_g_mean',
+                                                  }
+        config.astrometryRefObjLoader.anyFilterMapsToThis = None
```
parejkoj, line 76: Note: this entire jointcal config block can go away once I merge DM-29008 (which should happen today).

pipelines/HSC/RC2.yaml
```diff
@@ -0,0 +1,379 @@
...
+    # would fix this, and it's not a huge loss because we haven't actually
+    # gotten jointcal running in production on anything other than HSC (it
+    # might just work, but we certainly can't claim it's supported).
+    class: lsst.jointcal.JointcalTask
+    config:
+      # Disable photometry because we don't need it (FGCM takes care of that).
+      # It would be better to disable this via config (DM-29885), but that's
+      # more refactoring we haven't yet done.
+      doPhotometry: false
+      python: |
```
parejkoj, line 22: Note: this entire jointcal config block, except for the `doPhotometry` line just above can go away once I merge DM-29008 (which should happen today).

TallJimbo: After removing this block (I also removed the `doPhotometry` line, because of DM-29885), the expanded config from the pipeline on this branch differs from the expansion of `obs_subaru` `main` pipeline in two ways:

 - `main` has `config.colorterms.data` populated, while this branch does not.
 - `main` has `config.photometryRefObjLoader.filterMap` populated, while this branch doe not.

I'm fairly certain those don't matter as long as `doPhotometry` is `False`, but if they're necessary to make `doPhotometry=True` work on HSC, it seems appropriate to include them so turning on photometry just involves setting `doPhotometry=True`, and not populating these dicts.  Or is it no longer necessary to populated them anyway, due to the `anyFilterMapsToThis` stuff?  Is there any reason to still include them in the defaults, if that's the case?

(also CC @laurenam)

parejkoj: Ah, good point! I removed the jointcal photometry filterMap from obs_subaru, but you're correct that we want it so there's just a single step to enable jointcal photometry fitting for HSC. I think this is a case where it's more appropriate to have the config it in the obs package, since you want it whether you're running the DRP pipeline or just jointcal itself.

laurenam: Yeah, I think that makes sense.  I’m a bit torn as I’m not a fan of having configs set that aren’t being used, but I see where having these set could save some pain for anyone wanting to turn photometry on in `jointcal`.  I suppose a caveat is that there needs to be colorterms appropriate for the refcat used for the photometry, but it looks like the default for `jointcal` is PS1, so all is good there.

TallJimbo: Ok, I've pushed a commit to the `obs_subaru` branch of this ticket that puts the `colorterms.py` and `filterMap.py` config loads back into the `jointcal.py`.  I've also set
```
config.applyColorTerms = True
```
there, which is actually different from the current HSC pipeline on `main` (i.e. on `main`, `applyColorTerms=False` for jointcal in Gen3).  Can either of you confirm that this is a no-op when photometry is not being run, and not a problem?

laurenam: Looks like you’ll just get a warning:
https://github.com/lsst/jointcal/blob/main/python/lsst/jointcal/jointcal.py#L589-L592

parejkoj: Correct: jointcal does nothing with colorterms when `doPhotometry=False`.

pipelines/DECam/Merian.yaml
```diff
@@ -0,0 +1,195 @@
+description: |
+  The DRP pipeline specialized for the DECam instrument, developed against the
+  Merian dataset.
```
mrawls, line 3: I suggest calling this `DRP.yaml` unless there is a strong reason not to. If I am a naive new pipelines user looking for the correct DRP pipeline to run on my DECam data, I will not expect something called `Merian.yaml` to be it.

ingredients/LSST/DC2/DRP.yaml
```diff
@@ -0,0 +1,221 @@
+description: DRP specialized for ImSim-DC2 data
```
mrawls, line 1: I am confused about the directory naming. In `ap_pipe` we have `DarkEnergyCamera`, `HyperSuprimeCam`, and `LsstCamImSim` as the three `pipelines` subdirectories. Shouldn't it be the same here in both `ingredients` and `pipelines`? I am also confused why "DC2" gets its own `DRP.yaml` in ingredients, but HSC and DECam do not.

pipelines/DECam/Merian.yaml
TallJimbo: The idea was that we _would_ have an `ingredients/DECam/DRP.yaml` someday, and this and other more concrete, specialized pipelines would mostly defer to that.  I wanted to get the directly-runnable pipelines in place before doing that refactoring on another ticket.  And I didn't want to have `pipelines/*/DRP.yaml` files because I only wanted to put things in `pipelines` that had actually been run on some dataset (and ideally either run regularly or run in some important production).  I'd rather have somebody who wants to run on new data start from the ingredients and think about what kind of processing they want, even if they end up _just_ inheriting from the ingredients pipeline.

But I think you're right that it really doesn't make sense to have just a single more-concrete pipeline file in this package that's mostly (and maybe wholly) full of quite general stuff.  And it's easy to do a lot better, by renaming this file to `ingredients/DECam/DRP.yaml` and making a new `pipelines/DECam/Merian.yaml` that just points to that - that's what I already did with the LSST/DC2 pipelines, after all.  I'll do at least that much.

I'm willing to consider going further, by also adding a `pipelines/DECam/DRP.yaml` that also just forwards to the one in ingredients; I think that's a win _if_ we end up having a single generic pipeline that can actually just work on a lot of random data.  I'm hesitant because this hasn't turned out to be the case for HSC, where our "best" pipeline is really quite RC2-specific, because the FGCM configuration is really quite dependent on the set of bands being processed together (and maybe some other properties of the dataset, but at least that).  I suppose this DECam pipeline doesn't _obviously_ have that problem, though, so maybe we should have a generic DRP.yaml until we have a concrete reason we shouldn't.  :man_shrugging:

ingredients/LSST/DC2/DRP.yaml
TallJimbo: Second question first, because I think it's mostly covered by the other thread: DC2 got its own `DRP.yaml` ingredients because it was easy, HSC didn't because it's hard and I wanted to punt any nontrivial refactoring to another ticket, and DECam didn't because I didn't notice how easy it would be, and so I'll go fix that now.

As for the names, the ones I've used here are the "short" names used in butler data IDs and collection names.  I do think those names are preferably almost everywhere the Python class names (which I gather is what ap_pipe is currently using).  The Python class names are constrained both by our coding conventions (camelcase! no acronyms!) and greater potential for name conflicts in a way the short names generally are not, and I think the short names are much more user-friendly.  So I'm hoping we could actually change `ap_pipe` on this one.

As an aside, the LSST/DC2 name is a little weird because - being simulations - the instrument isn't really a solid identifier for the data, because ImSim can fundamentally change between versions in a way physical instruments can't.  That's why `/repo/dc2` is a different repo entirely at NCSA, and "2.2i" (the simulation run version) is used in collection names there instead of just the instrument name.

pipelines/HSC/RC2.yaml
TallJimbo: Sounds like we should probably remove that warning, but I'll leave the pipeline here as it is, then.

pipelines/HSC/RC2.yaml
```diff
@@ -0,0 +1,340 @@
...
+tasks:
+  skyCorr:
+    class: lsst.pipe.drivers.skyCorrection.SkyCorrectionTask
+  jointcal:
+    # Running jointcal is something we want to do for all real-world
+    # instruments, so it doesn't belong in an HSC-specific pipeline, let alone
+    # an RC2-specific one.  We just haven't done the pipeline refactoring that
+    # would fix this, and it's not a huge loss because we haven't actually
+    # gotten jointcal running in production on anything other than HSC (it
+    # might just work, but we certainly can't claim it's supported).
```
parejkoj, line 15: I'd remove some of this statement. Jointcal should absolutely work in production on DECam and CFHT: I've been testing it on those datasets all along (until I had to remove the DECam tests because of lack of gen3 test data). It should also work on LSST, if we had any data (the simulated data isn't useful). I'd end the sentence at "...that would fix this.", without the paranthetical.

pipelines/HSC/ci_hsc.yaml
```diff
@@ -0,0 +1,45 @@
+description: The DRP pipeline specialized for the ci_hsc test dataset.
+instrument: lsst.obs.subaru.HyperSuprimeCam
+imports:
+  # In the long term, we should try to avoid importing anything from pipelines;
+  # better to refactor anything imported into ingredients.
+  location: "$DRP_PIPE_DIR/pipelines/HSC/RC2.yaml"
+  exclude:
+    - fgcm
+    # Don't run jointcal here...
```
parejkoj, line 9: Why this comment, but not have one for fgcm? Maybe "Don't run jointcal or fgcmcal because we don't have enough data in this dataset."

pipelines/DECam/Merian.yaml
parejkoj: It's probably not worth doing on this ticket, but a future refactoring should absolutely pull out those RC2-specific FGCM configs, so that the generic parts of RC2 (lots of faro, looks like?) can be moved to a more generic place.

mrawls: I think making something that's obviously DECam-flavored `DRP.yaml` in either ingredients or pipelines is sufficient for now. I'd love the same for all cameras, just to have some vague idea of what a functional/basic DRP pipeline is for each camera. It's the kind of thing I envision new team members and folks on community asking for when they "just want to try processing some data."

ingredients/LSST/DC2/DRP.yaml
mrawls: I forget exactly why we chose the Python class names for `ap_pipe`, but I think at least part of the reason was because I kept having to look up what `instrument` the butler was governor-haranguing me to specify, and it seemed the most canonical choice. Anyway, I'm fine switching to shorter names for `HSC` and `DECam` in `ap_pipe`.

Regarding `DC2`. I enthusiastically approve of keeping fake data wholly separate from real data (separate repos, as you say). But with this pipeline and ingredient directory structure, in the future we'll have `LSST/LsstCam` alongside `LSST/DC2`? It seems wrong to have both fall under `LSST` when one is fake and one is not, and when they would never ever share a butler repo.

pipelines/HSC/RC2.yaml
```diff
@@ -0,0 +1,340 @@
+description: The DRP pipeline specialized for the HSC RC2 dataset.
```
kfindeisen, line 1: Consider including pipeline name in files (e.g., DRP-RC2), because there are pipelines other than DRP in `drp_pipe` (e.g., `QaTractTables`).

kfindeisen (approved):
Looks good; most of my questions are about process (i.e., having a representative and understandable commit history, clarifying which files have been copied and which ones are new) rather than anything fundamental.

I have one general question about how you plan to deal with future changes to the `obs_*` config files. You said yesterday that you don't want to phase out the configs until we've taken care of everything else Gen 2, but these new pipelines are already taking over some of the configs (e.g., for `jointcal`). Are you planning to resynchronize the changes at a later date?

ingredients/LSST/DC2/DRP.yaml
kfindeisen: @TallJimbo, is the fact that these should be short names written down anywhere other than RFC-775? I believe that standardization is much more important here than what the standard actually is.

As for DC2 pipelines, I kind of like the current arrangement, since I think the only reason we care about simulated data is as a proxy for real data, so the processing should be as similar as possible. Even if the ImSim "instrument" needs to be versioned, it's still an LSST instrument.

pipelines/DECam/Merian.yaml
kfindeisen: I agree with @mrawls that it would be good to have a generic-ish pipeline that new people can start from, even if it's not perfect. I'd also suggest calling this file something like `DRP-Merian`, just to make it clear that it is a specialization of DRP rather than an auxiliary pipeline (see my comment below about `RC2.yaml`).

README.rst
```diff
@@ -4,4 +4,9 @@ drp_pipe
 
 ``drp_pipe`` is a package in the `LSST Science Pipelines <https://pipelines.lsst.io>`_.
 
-.. Add a brief (few sentence) description of what this package provides.
+This package contains the LSST Data Management Data Release Pipeline (DRP).
+
+``drp_pipe`` contains the pipeline files and related configurations to process raw images that have been ingested into a Butler repository with corresponding calibration products.
+The DRP pipeline produces exposures with astrometric and photometric calibrations, coadds, source catalogs, and visit-level summary catalogs.
+
+For in-depth documentation, see the doc directory.
```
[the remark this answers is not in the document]
kfindeisen: Actually, can you add `drp_pipe` to `pipelines_lsst_io` on this ticket, instead of on DM-33034?

ups/drp_pipe.table
```diff
@@ -3,6 +3,15 @@
...
 setupRequired(base)
+setupRequired(obs_subaru)
+setupRequired(obs_lsst)
+setupRequired(obs_decam)
+setupRequired(obs_subaru)
+setupRequired(pipe_tasks)
+setupRequired(meas_algorithms)
+setupRequired(jointcal)
+setupRequried(fgcmcal)
+setupRequired(ctrl_bps)
```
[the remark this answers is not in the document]
timj: Did we have a discussion as to why this needs to depend on ctrl_bps at all? Presumably the pipeline can run just fine without it -- you just can't process very much data. Can this specific line be setupOptional?

ingredients/LSST/DC2/DRP.yaml
TallJimbo: > is the fact that these should be short names written down anywhere other than RFC-775?

DMTN-167 (organization of and naming conventions for data repositories) does talk about using the short names there, but I think RFC-775 is the only document that talks about this particular case at all.  Documenting this convention somewhere seems like a good idea, though; best idea I have would be README files in the `pipelines` directories of `drp_pipe` and `ap_pipe`, but I'm certainly open to other ideas, too.

> But with this pipeline and ingredient directory structure, in the future we'll have LSST/LsstCam alongside LSST/DC2

Like @kfindeisen , I do think it's appropriate to include LSST in these names, but I think @mrawls makes a good point that `LSST/LsstCam` (or `LSST/LSSTCam` with the actual short name) is a bit awkward, and just `LSST` is a little awkward, too, since it makes the real-LSST pipelines siblings of the DC2 directory.  Maybe using "Rubin" is an easy way out, even if it _isn't_ part of the short name (maybe it should be, or should have been, if timing had been different):
```
DECam/*
HSC/*
Rubin/DC2/*   # DC2 pipelines
Rubin/LSST/*   # real LSST pipelines
```
Another option would be to use LSST in the prefix for the DC2 directory, without using a level of directory structure:
```
DECam/*
HSC/*
LSST-DC2/*   # DC2 pipelines
LSST/*   # real LSST pipelines
```
Any preference or other ideas?

kfindeisen: I'd like to avoid special-casing (it's hard for human pipeline users to figure out, and extra-hard for machines), so I don't really like either of these. I'd prefer:

```
    DECam/*
    HSC/*
    ImSim/*
    LSSTCam/*
```

That's assuming I got the short names right, of course...

pipelines/HSC/ci_hsc.yaml
TallJimbo: Just history copied over from the originals.  Will fix.

recipes/DECam/DRP.yaml
```diff
@@ -0,0 +1,16 @@
+description: DRP specialized for DECam
+instrument: lsst.obs.decam.DarkEnergyCamera
+imports:
+  - location: $PIPE_TASKS_DIR/pipelines/DRP.yaml
+    exclude:
+      - sourceTable
+      - objectTable
+      - isr
+  - location: $AP_PIPE/pipelines/RunIsrWithCrosstalk.yaml
```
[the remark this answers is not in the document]
TallJimbo: @leeskelvin , can you help me out with what happened here?  I think you may have committed the first version of this pipeline to drp_pipe, then committed a later one to obs_decam (when drp_pipe lingered on this branch for a while), and then I synced them up later.  Does that sound right?

recipes/DECam/DRP.yaml
```diff
@@ -0,0 +1,16 @@
...
+# This DRP pipeline uses an obs_decam specific ISR configuration that
+# requires crosstalk sources to be pre-generated via the
+# RunIsrForCrosstalkSources.yaml pipeline.  This applies overscan
+# correction and writes a new dataset that is then used as needed for
+# the inter-chip crosstalk solutions.  Unless a specific overscan
+# change is required, it may be best to generate these crosstalk
+# sources soon after ingest to allow them to be readily available.
```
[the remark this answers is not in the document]
TallJimbo: Copied from obs_decam (or maybe mis-copied), but I'll fix it as suggested.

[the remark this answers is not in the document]
TallJimbo: Actually it looks like this got fixed in a later sync to the version in obs_decam (but I also see some bad merge conflict resolution in the description string that now covers this topic, which I'll fix).

recipes/HSC/DRP.yaml
```diff
@@ -0,0 +1,380 @@
...
+      # jointcal warns if that's true but doPhotometry=False.  It would be
+      # better to disable this via config (DM-29885), but it's hard to make
+      # that compatible with the jointcal tests because jointcal depends on
+      # obs_subaru.
+      doPhotometry: false
+      applyColorTerms: false
+      python: |
+        # These overrides were copied from the tests/config/config-gen3.yaml file
+        # on 22 Sept 2021.  This is necessary for processing in containers to succeed
+        # because the build system strips out the test directories for space concerns.
```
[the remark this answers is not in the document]
TallJimbo: Happily these overrides are all gone now, along with this comment: tickets merged to master (and then synced to this branch) removed this hack.  IIRC, the situation was that the canonical jointcal configs for gen3 first landed in that configuration used to drive the jointcal unit tests, and it was mess trying to get those available to both the pipeline and the tests due to the direction package dependencies went and the need to not clobber the gen2 configs.

recipes/HSC/DRP.yaml
```diff
@@ -0,0 +1,380 @@
...
+        #
+        # TODO DM-29008: When that ticket is merged these overrides will no longer
+        # be needed at all since that will fix higher level issues with how config
+        # overrides are imported in general.
+        config.sourceSelector['science'].flags.bad = ['pixelFlags_edge',
```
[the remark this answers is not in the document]
TallJimbo: The content was present in obs_subaru prior to c1436cadac54fc2b24a4f868d19897c05ea25d6e

ups/drp_pipe.table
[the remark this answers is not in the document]
TallJimbo: Will add comments.

ctrl_bps was included because this package will eventually BPS config files (or includeable pieces of them), so it will have a dependency in a "a change to ctrl_bps [configuration] APIs may affect this package".  That was raised in the RFC but never _really_ got resolved; my preference may have won out only because I exhausted those who disagreed by giving them too much to read (see "Variant A" [here](https://jira.lsstcorp.org/browse/RFC-775?focusedCommentId=274849&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-274849), [here](https://jira.lsstcorp.org/browse/RFC-775?focusedCommentId=276579&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-276579), and in [the proposal on confluence](https://confluence.lsstcorp.org/display/~jbosch/Reorganizing+Top-Level+Packages+and+Pipelines)).

In any case, there is BPS configuration here now, so I'll remove the dependency for now and we can discuss again when that changes.

recipes/DECam/Merian.yaml
```diff
@@ -71,78 +71,30 @@ tasks:
...
 subsets:
   processCcd:
     subset:
     - isr
     - characterizeImage
     - calibrate
-    description: 'An alias of ProcessCcd to use in higher-level pipelines.'
+    description: 'An alias of ProcessCcd to use in higher-level pipelines.
+      
+      '
```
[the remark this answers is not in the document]
TallJimbo: I think this is an artifact of round-tripping a string originally defined with `>` or `|` (and maybe some carelessness?) through pyyaml read-write.  I'm not sure how it got here originally, but it got cleaned up on the latter "YAML formatting and linting" commit.

recipes/DECam/Merian.yaml
```diff
@@ -1,82 +1,47 @@
...
     class: lsst.ip.isr.IsrTask
     config:
-    - connections.ccdExposure: raw
-      connections.outputExposure: overscanRaw
-      doBias: false
+    - connections.outputExposure: overscanRaw
       doOverscan: true
       doAssembleCcd: false
+      doBias: false
       doVariance: false
```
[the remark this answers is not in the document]
TallJimbo: This boils down to the same question I punted to @leeskelvin on another thread.

I'm beginning to think that the best way to fix the history here would be to try to squash a lot of it, but the "transferred this over" commits are interleaved with the "cleaned this up" commits, so all but the initial transfers involved conflict resolution, and hence changes.

ingredients/DRP+fakes.yaml
```diff
@@ -1,131 +0,0 @@
...
-      connections.outputCoaddName: "fakes_deep"
-  healSparsePropertyMaps:
-    class: lsst.pipe.tasks.healSparseMapping.HealSparsePropertyMapTask
-    config:
-      connections.coaddName: "fakes_deep"
-      connections.calexpType: "fakes_"
-  consolidateHealSparsePropertyMaps:
-    class: lsst.pipe.tasks.healSparseMapping.ConsolidateHealSparsePropertyMapTask
-    config:
-      connections.coaddName: "fakes_deep"
```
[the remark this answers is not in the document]
TallJimbo: I'm pretty sure it's already broken, at least because it doesn't have the `processCcdWithFakes` -> `singleFrameWithFakes` label rename that HSC/RC2+fakes has; I took that as an indication that there are probably other problems, but if you want me to make that change and include it while hoping for the best, I can.

ups/drp_pipe.table
```diff
@@ -1,17 +1,17 @@
...
 setupRequired(obs_subaru)
+
+# These packages provide the tasks directly referenced by the pipelines
+# in this package.
+setupRequired(ip_isr)
 setupRequired(pipe_tasks)
-setupRequired(meas_algorithms)
+setupRequired(meas_base)
 setupRequired(jointcal)
-setupRequried(fgcmcal)
```
kfindeisen: Whoops! 😆

ingredients/LSST/DC2/DRP.yaml
TallJimbo: Done; the short name for ImSim is actually `LSSTCam-imSim`, and that's what I've used.

TallJimbo:
@kfindeisen , thanks for the careful review and good suggestions, and apologize for suggesting looking at the original commit history - I think it's pretty clear I should have squashed a lot more than I did originally, because in this case trying to make fine-grained commits that showed exactly what I actually did completely obfuscated the actual transfers.

Disentangling that completely - having one transfer commit, followed by nice, fine-grained cleanup and reorganization commits - would have meant starting over from the current state of the origin repositories, and that doesn't seem worthwhile.  So instead I've done a lot of squashing, with commit messages that summarize what happened but at least don't lie, as the previous ones did.  I'm not positive I've addressed all comments, but I'm going to go through now and resolve those that I have resolved (including those purely about history), so I can more easily track what remains.  If you still have concerns about the git history and traceability, let's discuss those generally, rather than on any of those particular line-comment threads.

> I have one general question about how you plan to deal with future changes to the obs_* config files. You said yesterday that you don't want to phase out the configs until we've taken care of everything else Gen 2, but these new pipelines are already taking over some of the configs (e.g., for jointcal). Are you planning to resynchronize the changes at a later date?

There should be no such changes on this ticket anymore - the appearance of such moves was another artifact of the confusing history (I was just tracking obs_subaru's pipeline, but it was a moving target, and the config->pipeline->config moves there were workarounds for a specific set of problems).

recipes/DECam/DRP.yaml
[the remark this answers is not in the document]
TallJimbo: Nevermind, Lee - I've squashed this problem away.

ups/drp_pipe.table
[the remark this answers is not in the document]
TallJimbo: Ended up doing more than just adding comments - the dependency list has been pretty much completely rewritten, to include:
 - the obs packages whose configs these pipelines will load when expanded
 - the packages that define the PipelineTasks referenced here.

pipelines/DECam/Merian.yaml
```diff
@@ -81,11 +86,11 @@ tasks:
...
       matchingKernelSize: 29
       doApplyExternalPhotoCalib: true
-      externalPhotoCalibName: 'jointcal'
+      externalPhotoCalibName: "jointcal"
       doApplyExternalSkyWcs: true
       modelPsf.defaultFwhm: 7.7
-      warpAndPsfMatch.warp.warpingKernelName: 'lanczos5'
-      coaddPsf.warpingKernelName: 'lanczos5'
+      warpAndPsfMatch.warp.warpingKernelName: "lanczos5"
+      coaddPsf.warpingKernelName: "lanczos5"
```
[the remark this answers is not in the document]
TallJimbo: I bet they were copied from Python config files and not edited enough for consistency.  Will fix.

pipelines/LSST/DC2/ci_imsim.yaml
```diff
@@ -0,0 +1,26 @@
...
+imports:
+  - $DRP_PIPE_DIR/ingredients/LSST/DC2/DRP.yaml
+tasks:
+  deblend:
+    class: lsst.pipe.tasks.deblendCoaddSourcesPipeline.DeblendCoaddSourcesMultiTask
+    # Significantly limit the number of sources to deblend, which saves time
+    # in both deblending and measurement
+    config:
+      # Only run the deblender on a small subset of blended parents
+      multibandDeblend.useCiLimits: true
```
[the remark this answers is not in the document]
TallJimbo: This is a mess over in `ci_imsim`: the options do indeed have a default of `False` in the argument parser over in `ci_imsim_run.py`, but then the `ups/eupspkg.cfg.sh` script that runs it in CI explicitly passes `--config-limit-deblend` on the command-line making them `True`, so that's the behavior most people are used to.

I don't think I can fix this without fixing the inconsistent behavior of `ci_imsim` itself, and I don't want to do that here; I'll open a ticket, because that's a problem in its own right.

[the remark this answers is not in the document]
TallJimbo: [DM-33459](https://jira.lsstcorp.org/browse/DM-33450)

recipes/DECam/DRP.yaml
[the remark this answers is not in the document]
leeskelvin: Apologies @TallJimbo, I just saw your earlier message! Yes, as long as the `DRP.yaml` in `ingredients/DECam/DRP.yaml` matches that in `$OBS_DECAM_DIR/pipelines/DRP.yaml`, then this looks okay to me.

ingredients/LSST/DC2/DRP.yaml
mrawls: This is good by me. I made DM-33453 to update the ap_pipe pipelines directory structures accordingly.

PR #3, opened 2022-02-01
DM-26987: Update filtering of matched catalog to operate before creating GroupView

jeffcarlin:
The new tasks in `faro` for matched catalog creation also needed to be added to the default DRP.yaml pipelines here.

### DIFF
--- a/ingredients/LSSTCam-imSim/DRP.yaml
+++ b/ingredients/LSSTCam-imSim/DRP.yaml
@@ -110,6 +110,9 @@ subsets:
       - matchCatalogsTract
       - matchCatalogsPatch
       - matchCatalogsPatchMultiBand
+      - matchCatalogsTractMag17to21p5
+      - matchCatalogsTractStarsSNR5to80
+      - matchCatalogsTractGxsSNR5to80
       - PA1
       - PF1_design
       - AM1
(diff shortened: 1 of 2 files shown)


jeffcarlin:
[Jenkins build](https://ci.lsst.codes/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/35818/pipeline/) gives a green light.

ingredients/LSSTCam-imSim/DRP.yaml
```diff
@@ -110,6 +110,9 @@ subsets:
       - matchCatalogsTract
```
ctslater, line 110: Aren't the new tasks meant to replace this?

jeffcarlin: That one stays around as a "default" (with a somewhat conservative S/N>50 cut).

ctslater: Is anything using that? I'm a little hesitant about creating datasets that aren't being used.

jeffcarlin: Yeah, the photometric repeatability metrics (PA1, PF1) use it.

PR #5, opened 2022-02-14
DM-33279: Add IsolatedStarAssociationTask.

erykoff:
I am not entirely sure where to put the task in terms of the steps.  This should be run per-tract but prior to fgcmcal when fgcmcal supports it.


fred3m (commented):
Everything looks fine here, I'm just curious why you didn't add this task to any of the other pipeliens, like DECa, LSST-ImSim, or any of the other HSC pipelines.

erykoff:
I added the task to the generic `DRP.yaml` which applies to everything, and then to particular steps in the `HSC` pipeline.  That might have been premature since where it should go in the RC processing depends on the exact format of the next `FinalizeCharacterizationTask`.

PR #7, opened 2022-02-28
DM-32141: Get SFM working for LATISS

mfisherlevine:


pipelines/LATISS/DRP.yaml
```diff
@@ -0,0 +1,14 @@
...
+instrument: lsst.obs.lsst.Latiss
+imports:
+  - location: $DRP_PIPE_DIR/ingredients/DRP.yaml
+    include:
+      - processCcd
+      - writeSourceTable
+      - transformSourceTable
+
+subsets:
+  sourceTable:
```
parejkoj, line 11: Why do you have this as a separate subset?

mfisherlevine: Yeah, I don't actually know tbh. I guess because it's non-standard, as that usually has other steps, and I wanted to make the one for this not call the broken bits if people did use it alone, but it's possibly pointless... I'm still quite new to this stuff. Should I just remove that altogether do you think?

parejkoj: Subsets exist so you can use them in a pipetask call. Do you have a usecase for building the consolidated source tables in a separate command? Otherwise, no reason to have this.

mfisherlevine: Only because we used to run the first three steps first, alone, and so did sometimes call this on its own while we were building things, but yeah, I think it can go now 👍

edennihy: I dropped this subset and added a few other imports needed for complete singleFrame processing and multiVisit processing as part of 604583d

PR #10, opened 2022-03-14
DM-33963: refactor pipeline content and add pipelines for gathering resource usage statistics

TallJimbo:


TallJimbo:
Closing this for now; the changes here would have made the existing pipelines too fragile, and a separate pipeline yields extremely slow QG generation.  I'm working around this with a custom QG-gen script in analysis_drp for now.

PR #13, opened 2022-03-18
DM-33982: Add external calibration support to forced ccd photometry.

erykoff:


pipelines/HSC/DRP-RC2+fakes.yaml
```diff
@@ -117,6 +117,8 @@ tasks:
     class: lsst.meas.base.forcedPhotCcd.ForcedPhotCcdTask
     config:
       connections.inputCoaddName: "fakes_deep"
+      doApplyExternalPhotoCalib: true
```
yalsayyad, line 120: Delete these 2 lines. They are *already* true, as set here: https://github.com/lsst/obs_subaru/blob/main/config/forcedPhotCcd.py

pipelines/HSC/DRP-RC2.yaml
```diff
@@ -89,10 +89,23 @@ tasks:
   forcedPhotCcd:
     class: lsst.meas.base.forcedPhotCcd.ForcedPhotCcdTask
     config:
-      # The application of external calibrations doesn't work with Gen3 now (DM-17062)
-      doApplyExternalPhotoCalib: false
-      doApplyExternalSkyWcs: false
-      doApplySkyCorr: false
+      doApplyExternalPhotoCalib: true
```
yalsayyad, line 92: Delete these. They are already true here set here: https://github.com/lsst/obs_subaru/blob/main/config/forcedPhotCcd.py

pipelines/HSC/DRP-RC2.yaml
```diff
@@ -89,10 +89,23 @@ tasks:
...
+      doApplyExternalSkyWcs: true
+  forcedPhotDiffim:
+    class: lsst.meas.base.ForcedPhotCcdTask
+    config:
+      doApplyExternalPhotoCalib: true
```
yalsayyad, line 97: Delete these. They are already true here set here: https://github.com/lsst/obs_subaru/blob/main/config/forcedPhotCcd.py

pipelines/HSC/DRP-RC2.yaml
```diff
@@ -89,10 +89,23 @@ tasks:
...
+      doApplyExternalSkyWcs: true
+  forcedPhotCcdOnDiaObjects:
+    class: lsst.meas.base.ForcedPhotCcdFromDataFrameTask
```
yalsayyad, line 100: Delete these overrides and friends below, and make an `obs_subaru/config/forcedPhotCcdFromDataFrame.py`. 

Applying jointcal/FGCM during diffim  is still a TODO, so making these true is kinda inconsistent (if there'a 1-observation diaObject it won't be force photometered at the  exact xy of its centroid as measured on that one diffim).  I could go either way, since we already know that what we have isn't the best that we do anyway, and will be fixed in April. You're welcome to make a note to remind me to do this on  DM-31292 if you don't want to make a `obs_subaru/config/forcedPhotCcdFromDataFrame.py` now.    

But either way, these overrides here go away.

yalsayyad (approved):

pipelines/HSC/DRP-RC2.yaml
erykoff: I don't exactly follow your logic, I think we should discuss later.  In the meantime I'll comment on DM-31292.

PR #17, opened 2022-04-20
DM-34489: Add faro flux stat task into RC2/DC2 pipelines

leeskelvin:


jeffcarlin (approved):
Looks good -- we'll want to check to be sure this ran OK in `verify_drp_metrics` the day after you merge (hopefully it runs fine and the the sky object metrics will be dispatched to squash!).

PR #21, opened 2022-05-16
DM-34809: Add isol. star assoc. / finalize char. to DECam

leeskelvin:


ingredients/DECam/DRP.yaml
```diff
@@ -120,6 +120,11 @@ tasks:
         config.badMaskPlanes += ["SUSPECT"]
         from lsst.pipe.tasks.selectImages import PsfWcsSelectImagesTask
         config.select.retarget(PsfWcsSelectImagesTask)
+  isolatedStarAssociation:
+    class: lsst.pipe.tasks.isolatedStarAssociation.IsolatedStarAssociationTask
+    config:
+      python:
+        config.band_order += ["N708", "N540"]
```
erykoff, line 127: Just clarifying that it's okay if the narrow bands are lowest in the matching priority list?

erykoff (approved):

ingredients/DECam/DRP.yaml
leeskelvin: During my test runs, it does complete successfully, yes. The DECam Merian data does not have any broad-bands, so it cycles through this list until it hits the N708 band.

PR #27, opened 2022-06-28
add BPS configuration for BPS clusters

PaulPrice:
Tested on ci_hsc.


TallJimbo:
Thanks, @PaulPrice - it might take me a week or so to get to it, because this is a particularly busy one and there's a closely related ticket branch from @brendal4 that I'd like to think about at the same time.  But if I seem to have forgotten entirely (say, by end of next week), please remind me.

PR #32, opened 2022-07-18
DM-35060: Use piff for characterizeImage in DRP

parejkoj:


TallJimbo (approved):
What's going on with `yamllint` here is that its default (which we use) for the [indent-sequences](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.indentation) rule is "consistent", which means it doesn't care whether you indent sequences relative to their parent or not, as long as you do it consistently within a file.  And your additions (which are, I suspect, now the first sequences in the file) do not indent, while all later ones do.

PR #47, opened 2023-01-04
DM-35207: update visit summaries and use them downstream

TallJimbo:


laurenam (approved):
This one is admittedly the hardest to wrap one's head around...but it does all look good to me.

PR #52, opened 2023-01-31
DM-37792: Add analysis_drp/analysis_tools tasks to DC2/imsim pipeline(s)

taranu:


ingredients/LSSTCam-imSim/DRP.yaml
```diff
@@ -207,6 +207,20 @@ subsets:
...
       in the data query.
+  step8:
+    subset:
+      - analyzeObjectTableCore
+      - analyzeObjectTableSurveyCore
+      - catalogMatchTract
+      - refCatObjectTract
+      - diff_matched_analysis
+    description: |
+      Core (most important), coadd-level plots/metrics from analysis_tools.
```
yalsayyad, line 218: All the other steps have instructions on what they can or cannot put in the data query. Make it clear that as long as "analyzeObjectTableSurveyCore" is here, that this is a global task and they may not put anything in the data query or split up into tracts

PR #54, opened 2023-02-14
DM-37984: Move rc2_subset DRP pipeline definitions to drp_pipe

leeskelvin:


pipelines/HSC/DRP-RC2_subset.yaml
```diff
@@ -0,0 +1,427 @@
...
+  nightlyStep2b:
+    subset:
+      - jointcal
+    description: >
+      Per-tract tasks.
+      jointcal should be run with explicit 'tract' constraints essentially all the
+      time, because otherwise quanta will be created for jobs with only partial
+      visit coverage.
+      isolatedStarAssociation can be run with or without 'tract' constraints.
+  nightlyStep2cde:
```
jeffcarlin, line 305: Does it work to include FGCM in the same step with the other tasks? In rc2_subset, FGCM has been separate because it explicitly has to be run on a single processor. Is that true of the other tasks here, or would it improve performance to keep them separate?

jeffcarlin: See the shell script that executes rc2_subset processing at https://github.com/lsst/rc2_subset/blob/bf14286d29ed909cd0b0949231dc2b19725cb8d1/bin/measureHscRC2Metrics.sh

pipelines/HSC/DRP-RC2_subset.yaml
```diff
@@ -0,0 +1,427 @@
...
+      'tract' or 'patch' as part of the data ID expression if all
+      reference catalogs or diffIm templates that cover these
+      detector-level quanta are desired.
+  nightlyStep5:
+    subset:
+      - nsrcMeasVisit
+      - TE1
+      - TE2
+      - TE3
+      - TE4
```
jeffcarlin, line 401: See note above - TE3 and TE4 are per-visit tasks, so should be in Step2cde?

pipelines/HSC/DRP-RC2_subset.yaml
```diff
@@ -0,0 +1,427 @@
...
+    description: >
+      Tasks that can be run together, but only after the 'step1', 'step2'
+      and 'step3' subsets
+      These detector-level tasks should not be run with
+      'tract' or 'patch' as part of the data ID expression if all
+      reference catalogs or diffIm templates that cover these
+      detector-level quanta are desired.
+  nightlyStep5:
+    subset:
+      - nsrcMeasVisit
```
jeffcarlin, line 397: This is a per-visit task, so should be in Step2cde?

jeffcarlin (commented):
A few small comments, but mostly looks good. Note that we'll also need to update the shell script in rc2_subset that executes the processing.

pipelines/HSC/DRP-RC2_subset.yaml
leeskelvin: Thanks Jeff. Following a conversation with @erykoff, I've decided to revert this change and maintain the three FGCM tasks in their own `nightlyStep3c` subset, with the remainder continuing on as before in `nightlyStep2d`. To that end, the only actual pipeline changes should now be the addition of `makeVisitTable` and `makeCcdVisitTable` into `nightlyStep2d`. Other proposed changes, such as moving `isolatedStarAssociation` from `nightlyStep2b` into `nightlyStep2a` (as in `DRP-RC2.yaml` have also been dropped).

pipelines/HSC/DRP-RC2_subset.yaml
```diff
@@ -0,0 +1,427 @@
...
+    subset:
+      - jointcal
+    description: >
+      Per-tract tasks.
+      jointcal should be run with explicit 'tract' constraints essentially all the
+      time, because otherwise quanta will be created for jobs with only partial
+      visit coverage.
+      isolatedStarAssociation can be run with or without 'tract' constraints.
+  nightlyStep2cde:
+    subset:
```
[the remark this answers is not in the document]
leeskelvin: These have been moved back to their former home in `nightlyStep2d`.

pipelines/HSC/DRP-RC2_subset.yaml
leeskelvin: These have been moved back to their former home in `nightlyStep2d`.

pipelines/HSC/DRP-RC2_subset.yaml
leeskelvin: This has been moved back to its former home in `nightlyStep2d`.

PR #64, opened 2023-05-09
DM-38927: Give requestMemory a refresh per w13 and w15 test-med-1 and RC2 reruns 

yalsayyad:


bps/resources/HSC/DRP-RC2.yaml
```diff
@@ -42,21 +36,21 @@ pipetask:
   healSparsePropertyMaps:
```
eigerx, line 36: I see that `templateGen` above requested 5000 and used 5237 of memory; do we want to edit that?

eigerx (commented):
I was looking for the increased memory in calibrate as well but I see that that issue is getting addressed in DM-39101. 

Did we miss `templateGen` ?

PR #83, opened 2023-09-14
DM-40555: Refactor LATISS photometric processing pipeline to add fgcmcal

erykoff:


hlin730 (approved):
Looks good, see comments in https://jira.lsstcorp.org/browse/DM-40555

PR #84, opened 2023-09-25
DM-35784: Update default truth match configuration

taranu:


pipelines/_ingredients/LSSTCam-imSim/DRP.yaml
```diff
@@ -69,7 +69,7 @@ tasks:
       python: |
         from lsst.pipe.tasks.diff_matched_tract_catalog import MatchedCatalogFluxesConfig
         columns_flux = {}
-        for band in 'ugrizy':
+        for band in ("u", "g", "r", "i", "z", "y"):
```
laurenam, line 66: If this is preferable, do you want to do it at lines 23 & 24 above?
```
        config.astromRefObjLoader.filterMap = {band: 'lsst_%s_smeared' % (band) for band in 'ugrizy'};
        config.photoRefObjLoader.filterMap = {band: 'lsst_%s_smeared' % (band) for band in 'ugrizy'};
```

pipelines/_ingredients/LSSTCam-imSim/DRP.yaml
```diff
@@ -35,31 +35,24 @@ tasks:
         # Target settings are likely common to all object tables
         from lsst.pipe.tasks.match_tract_catalog_probabilistic import MatchTractCatalogProbabilisticTask
         config.match_tract_catalog.retarget(MatchTractCatalogProbabilisticTask)
-        config.match_tract_catalog.columns_ref_flux = [
-            'flux_u', 'flux_g', 'flux_r',
-            'flux_i', 'flux_z', 'flux_y',
+        bands_match = ("u", "g", "r", "i", "z", "y")
+        # Bands to match aperture fluxes on as a fallback if cModel failed
+        bands_fallback = ("r", "i")
```
laurenam, line 40: Why `r` & `i`?

pipelines/_ingredients/LSSTCam-imSim/DRP.yaml
```diff
@@ -35,31 +35,24 @@ tasks:
...
+        config.match_tract_catalog.columns_target_err = [f"{col}Err" for col in columns_meas]
+        config.match_tract_catalog.coord_format.coords_ref_to_convert = {"ra": "x", "dec": "y"}
         # Might need adjusting for different survey depths
         config.match_tract_catalog.mag_faintest_ref = 27.0
-        config.match_tract_catalog.columns_ref_copy = ['id', 'truth_type']
-        config.match_tract_catalog.columns_ref_select_true = ['is_unique_truth_entry']
-        config.match_tract_catalog.columns_target_copy = ['objectId']
+        config.match_tract_catalog.columns_ref_copy = ["id", "truth_type"]
+        config.match_tract_catalog.columns_ref_select_true = ["is_unique_truth_entry"]
+        config.match_tract_catalog.columns_target_copy = ["objectId"]
```
laurenam, line 55: This file has a full-on mix-n-match of single vs. double quotes.  Instead of updating these here, perhaps a separate commit to homogenize all those in the task overrides would be warranted?

laurenam (approved):

pipelines/_ingredients/LSSTCam-imSim/DRP.yaml
taranu: Better yet, I think I'll put the filterMap in parameters now that those can be referenced in python blocks.

pipelines/_ingredients/LSSTCam-imSim/DRP.yaml
taranu: I didn't want to include every band because then the aperture fluxes would be weighted as much as the CModel fluxes for matching (there is no way to weight each column differently as of yet). It seems unlikely there are many good detections with only u/g aperture fluxes. I suppose there could be z-band dropouts but this was meant to fix a problem with brighter stars.

laurenam: Got it...a comment in the file would be nice.

PR #87, opened 2023-10-25
DM-41235: Update test to use Butler factory method

andy-slac:


timj (approved):
Much better and we won't need to fix it again when registry disappears.

PR #90, opened 2023-11-01
DM-41503: Add selectDeepCoaddVisits to DRP pipelines to select by seeing.

erykoff:
This ticket adds a sensible default for seeing selection, plus a specific cut for LATISS processing.

### DIFF
--- a/pipelines/LATISS/DRP.yaml
+++ b/pipelines/LATISS/DRP.yaml
@@ -39,6 +39,13 @@ tasks:
       photo_calib_provider: "global"
       background_provider: "input_summary"
 
+  selectDeepCoaddVisits:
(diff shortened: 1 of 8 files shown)


pipelines/LATISS/DRP.yaml
```diff
@@ -32,12 +32,12 @@ tasks:
     config:
       connections.cycleNumber: 6
 
-  updateVisitSummary:
```
yalsayyad, line 35: Wait, did you really mean to delete updateVisitSummary?

erykoff: No.  No, I did not.  I have no idea how that happened.

PR #91, opened 2023-11-02
DM-41414: Update LATISS DRP step info for multiple-skymaps

leeskelvin:


hlin730 (approved):
Looks complete in noting that the skymap constraint is needed, in the relevant steps 2a through 4.

PR #93, opened 2023-11-09
DM-39808: Add dynamically generated RC2 and rc2_subset deepCoadd source injection pipelines to drp_pipe.

jtmccann:


SConstruct
```diff
@@ -1,4 +1,46 @@
...
+    ),
+    source=os.path.join(PKG_ROOT, "pipelines", "HSC", "DRP-RC2_subset.yaml"),
+    action=" ".join(
+        [
+            libraryLoaderEnvironment(),
+            "make_injection_pipeline -t deepCoadd -r $SOURCE -f $TARGET --overwrite",
+        ]
+    ),
+)
+RC2_injected_deepCoadd = env.Command(
```
arunkannawadi, line 34: Is there a reason why one is `rc2` and the other is `RC2`? I think both of these should just be `RC2`.

pipelines/HSC/DRP-RC2_subset+injected_deepCoadd.yaml
```diff
@@ -0,0 +1,2764 @@
+description: The DRP pipeline specialized for rc2_subset processing in jenkins and
+  tutorials
```
arunkannawadi, line 2: Same comments as mentioned in the other dynamically generated pipeline.

arunkannawadi (approved):

pipelines/HSC/DRP-RC2+injected_deepCoadd.yaml
```diff
@@ -0,0 +1,2392 @@
+description: The DRP pipeline specialized for the HSC RC2 dataset.
```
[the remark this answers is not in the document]
jtmccann: I can talk to Lee about what it might take to get a description feature like this.

Thanks for that catch, I will remove them on my next commit!

SConstruct
leeskelvin, line 34: As a variable, this should probably have been `rc2_injected_deepCoadd`. Apologies for not catching it before the merge. It's not a serious issue, but something one of us can address in a future ticket next time we're poking around in this package.

PR #95, opened 2023-11-30
DM-41881: Decrease nBins for LATISS focal plane plots in analysis_tools

laurenam:


jrmullaney (approved):
I've confirmed that this works as expected.

PR #99, opened 2023-12-12
Update LATISS DRP.yaml to move 3 *SourceTable tasks

hlin730:
Update LATISS DRP.yaml to move the 3 tasks writeRecalibratedSourceTable, transformSourceTable (both in step4), and consolidateSourceTable (in step6) to step2bcde.   This is in response to comments by Lauren MacArthur to run these tasks as early as possible so outputs are available for QA scripts, and by Clare Saunders that the sourceTable_visits outputs of consolidateSourceTable are needed by the analysis tasks in step3c.


laurenam (approved):
LGTM

PR #102, opened 2023-12-19
DM-42152: Update RC2 step8 description dataquery constraints

leeskelvin:


TallJimbo (approved):
Certainly an improvement, but do you happen to know if this is a place where a `band` constraint might be needed as well?  I know there are some analysis tasks that read the object table (which does not have `band` in its data ID) but have outputs with `band` in their data IDs, and those need to be constrained.

leeskelvin:
I think it's certainly possible to add a `band` constraint (if you're only interested in results from a particular band), but not required. I can add mention of that into the text, if helpful?

TallJimbo:
I was worried specifically about the tasks (sorry, I don't remember which) that attempt to run over all bands in the database if you don't limit them.  If that's not a problem here, no need to mention band in the docstring.

leeskelvin:
The bands which get run for analysis tools tasks are configurable for each task, so I think that's typically where you want to apply that logic. Having chatted with Nate about this just now, you certainly can use a band constraint for various step8 tasks, but it may cause problems or lead to unexpected behavior.

PR #103, opened 2023-12-22
DM-42307: Separate clustering.yaml for DC2 and RC2 cases

brianyanny:
- visit_detector (step1) clustering requires distinct pipetask lists for
- DC2 (LSSTCam-imsim) and RC2 (HSC) cases


cmsaunders (approved):
This looks good to me, but I'm not sure that we need a special HSC clustering file if it's not changing anything. I had originally thought the `visit_detector` from the general clustering file should go into the HSC one, but I think that's actually going to be the configuration for most of the processing we do, and it shouldn't go into a specialized clustering file. I think you could leave out the HSC-specific file for now, but it's up to you.

PR #110, opened 2024-03-04
DM-43033: Add fgcmcal to ComCamSim ops rehearsal pipeline

erykoff:


pipelines/LSSTComCamSim/DRP-ops-rehearsal-3.yaml
```diff
@@ -9,6 +9,17 @@ imports:
   - $ANALYSIS_TOOLS_DIR/pipelines/matchedVisitQualityCore.yaml
   - $ANALYSIS_TOOLS_DIR/pipelines/visitQualityCore.yaml
 tasks:
+  fgcmFitCycle:
+    class: lsst.fgcmcal.fgcmFitCycle.FgcmFitCycleTask
+    config:
```
cmsaunders, line 14: Why not include these with the other configs in `obs_lsst`?

cmsaunders (approved):

pipelines/LSSTComCamSim/DRP-ops-rehearsal-3.yaml
erykoff: By habit these are more of a "pipeline" thing and not a "config" thing.  And if we define a starflat pipeline these are the most likely values to change.

PR #116, opened 2024-03-20
DM-43406: Remove analyzeObjectExtended from non-CI pipelines

taranu:


pipelines/HSC/DRP-ci_hsc.yaml
```diff
@@ -10,8 +10,10 @@ imports:
       - faro_visit
       - faro_matched
       - faro_tract
-  # We generally don't want all visit-level plots for non-CI runs
+  # We generally only want visit-level and coadd extended plots for CI runs
   - location: $ANALYSIS_TOOLS_DIR/pipelines/visitQualityCore.yaml
+  - location: $ANALYSIS_TOOLS_DIR/pipelines/coaddQualityExtended.yaml
```
sr525, line 15: Why don't we want coaddQualityCore here as well?

taranu: It's already included from the ingredients, these are just extra inclusions.

PR #127, opened 2024-05-30
DM-44232: Replace makeWarp with makeDirectWarp

arunkannawadi:
and makePsfMatchedWarp.

### DIFF
--- a/pipelines/_ingredients/DECam/DRP.yaml
+++ b/pipelines/_ingredients/DECam/DRP.yaml
@@ -46,21 +46,24 @@ tasks:
       background_provider: "input_summary"
       connections.photoCalibName: "jointcal"
       connections.skyWcsName: "jointcal"
-  makeWarp:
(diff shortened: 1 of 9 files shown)


pipelines/_ingredients/DECam/DRP.yaml
```diff
@@ -199,7 +202,8 @@ subsets:
     subset:
       - selectGoodSeeingVisits
       - selectDeepCoaddVisits
-      - makeWarp
+      - makeDirectWarp
+      - makePsfMatchedWarp
```
arunkannawadi, line 206: @leeskelvin I haven't tested that these changes that I am making to DRP pipeline are okay on DECam beyond CI passing. I've done more detailed tests in DM-45426 on HSC and DC2 datasets. Is there a coadd dataId I should run this on on DECam data to make sure that the coadds are largely the same even after I swap out the warping tasks?

PR #135, opened 2024-07-02
DM-45087: Remove step8 from HSC/ImSim and move its analysis tasks upstream 

yalsayyad:
Now that we're running RC2/DC2s with the sasquatch butler for all steps, we can move the analysis tasks to the step which creates their input data products.

Move source table consolidation to the global step2 stage, and add consolidateSourceTable to the per-visit sourceTable cluster for [writeRecalibrated|transform|consolidate]SourceTable.

### DIFF
--- a/pipelines/HSC/DRP-RC2.yaml
+++ b/pipelines/HSC/DRP-RC2.yaml
@@ -92,14 +92,7 @@ tasks:
   objectTableCoreRefCatMatchWholeSkyPlot:
     class: lsst.analysis.tools.tasks.WholeSkyAnalysisTask
     config:
-      connections.inputName: "objectTable_tract_gaia_dr3_20230707_match_metricsTable"
-      connections.outputName: "objectTableCoreRefCatMatch_wholeSky"
-      atools.wholeSkyMetric: WholeSkyPlotTool
       atools.wholeSkyMetric.bands: ["g", "r", "i", "z", "y"]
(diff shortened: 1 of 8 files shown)


pipelines/HSC/DRP-RC2.yaml
```diff
@@ -92,14 +92,7 @@ tasks:
   objectTableCoreRefCatMatchWholeSkyPlot:
```
jtmccann, line 92: you can remove all the configs expect "band" for matchedVisitCoreWholeSkyPlot and objectTableCoreWholeSkyPlot

pipelines/HSC/DRP-RC2_subset.yaml
```diff
@@ -158,14 +158,7 @@ tasks:
   objectTableCoreRefCatMatchWholeSkyPlot:
```
jtmccann, line 158: you can remove all the configs expect "band" for matchedVisitCoreWholeSkyPlot and objectTableCoreWholeSkyPlot

jtmccann: you can also do this in pipelines/HSC/DRP-ci_hsc.yaml

jtmccann (commented):
Other than a couple comments everything looks good

TallJimbo (approved):

PR #143, opened 2024-08-02
DM-45192: Reconfigure RA for real ComCam

mfisherlevine:


pipelines/_ingredients/LSSTComCam/DRP.yaml
```diff
@@ -0,0 +1,149 @@
...
+      atools.stellarAstrometricRepeatability2: AstrometricRelativeRepeatability
+      atools.stellarAstrometricRepeatability2.xValue: 2
+      atools.stellarAstrometricRepeatability2.process.calculateActions.rms.annulus: 20
+      atools.stellarAstrometricRepeatability3: AstrometricRelativeRepeatability
+      atools.stellarAstrometricRepeatability3.xValue: 3
+      atools.stellarAstrometricRepeatability3.process.calculateActions.rms.annulus: 200
+      atools.stellarAstrometricRepeatability3.process.calculateActions.rms.threshAD: 30
+      python: |
+        from lsst.analysis.tools.atools import *
+  analyzeObjectTableCore:
```
yalsayyad, line 62: delete  `analyzeObjectTableCore` block.

pipelines/_ingredients/LSSTComCam/DRP.yaml
```diff
@@ -0,0 +1,149 @@
...
+      ]
+      atools.wholeSkyMetric.keysWithBand: parameters.objectTableCoreKeysWithBand
+      python: |
+        from lsst.analysis.tools.atools import *
+  # These don't actually work now:
+  photometricMatchPreVisit:
+    class: lsst.analysis.tools.tasks.photometricCatalogMatch.PhotometricCatalogMatchVisitTask
+    config:
+      extraColumns: ["x", "y", "ap09Flux", "ap09FluxErr"]
+      connections.refCatalog: uw_stars_20240524
```
yalsayyad, line 132: All these refcats will need to change. You can delete any lines with "uw_stars" in them or leave them, but they'll need to be revisited on the next ticket.

pipelines/_ingredients/LSSTComCam/DRP.yaml
```diff
@@ -0,0 +1,149 @@
...
+        from lsst.analysis.tools.contexts import *
+  matchedVisitCoreWholeSkyPlot:
+    class: lsst.analysis.tools.tasks.WholeSkyAnalysisTask
+    config:
+      connections.inputName: "matchedVisitCore_metricsTable"
+      connections.outputName: "matchedVisitCore_wholeSky"
+      atools.wholeSkyMetric: WholeSkyPlotTool
+      atools.wholeSkyMetric.xLimits: [120.0, 305.0]
+      atools.wholeSkyMetric.yLimits: [-55.0, 10.0]
+      atools.wholeSkyMetric.bands: ["g", "r", "i"]
```
yalsayyad, line 103: Delete these xLimits, yLimits, and bands overrides. Here and in the next one.

yalsayyad (approved):

pipelines/_ingredients/LSSTComCam/DRP.yaml
mfisherlevine: Ah yes, good point, thanks. I've just done a find & replace on all those to change them to `gaia_dr3_20230707`, because that's at least a little less wrong - I think that's probably right for the astrometry bits, and means that we should then only need to change the photometry parts. Hopefully that's better than leaving them, anyway.

PR #145, opened 2024-08-05
DM-39168: Add task needed for proper motion in metrics

cmsaunders:


pipelines/HSC/DRP-ci_hsc.yaml
```diff
@@ -74,6 +74,18 @@ tasks:
       atools.wholeSkyMetric.keysWithBand: parameters.objectTableCoreKeysWithBand
       python: |
         from lsst.analysis.tools.atools import *
+  analyzeMatchedVisitCore:
+    class: lsst.analysis.tools.tasks.AssociatedSourcesTractAnalysisTask
+    config:
+      applyProperMotions: false
```
parejkoj, line 80: Why disable PM correction in HSC? Wouldn't that still improve any metrics that compare with a refcat?

cmsaunders: This is just for ci_hsc, where there's not enough data to run `gbdesAstrometricFit`. I added comments here and for the other tasks with this config set.

PR #146, opened 2024-08-07
DM-45651: Write pipelines for LSSTCam's on-sky processing

mfisherlevine:


pipelines/_ingredients/LSSTComCam/DRP.yaml
```diff
@@ -24,10 +24,8 @@ tasks:
     config:
       connections.goodVisits: goodSeeingVisits
       nVisitsMin: 12
-      # For simulation 2 there are 60 visits per region
-      # between 60399.362 and 60402.419.
-      minMJD: 60399.362
-      maxMJD: 60402.419
+      minMJD: 60399.362  # these need updating
```
yalsayyad, line 27: Whyyyy are these back? For ComCam on sky we're going to be using all the data taken to date to make templates for the next round of observations.

yalsayyad (approved):

pipelines/_ingredients/LSSTComCam/DRP.yaml
mfisherlevine: I just didn't know if they could just not exist at all, so left them in with a comment to update them, rather than removing something that could have been a required key, as far as I knew. I take it I should just delete these then?

yalsayyad: Yeah, my preference is to just delete them.

mfisherlevine: Done, and same for LSSTCam.

PR #150, opened 2024-08-22
DM-45886: Turn off calib application while we're working with sims

mfisherlevine:


erykoff (approved):
I look forward to blowing this away next week!
Also, these aren't really "slow" steps are they?

mfisherlevine:
I've improved the comment 🙂 When you update to the new task, as discussed, please just leave gain and overscan correction on and don't enable anything else, even though there will be calibs.

PR #155, opened 2024-09-13
DM-46327: Exclude FGCM tasks from LSSTCam/LSSTComCam

leeskelvin:


pipelines/_ingredients/LSSTCam/DRP.yaml
```diff
@@ -75,7 +75,8 @@ tasks:
     class: lsst.analysis.tools.tasks.refCatSourceAnalysis.RefCatSourceAnalysisTask
     config:
       # Only run metrics for analyzing the preSources:
-      connections.data: preSourceTable_visit_gaia_dr3_20230707_match_astrom
+      connections.data: preSourceTable_visit_uw_stars_20240524_match_astrom
```
cmsaunders, line 78: I think what was already here was correct. The `uw_stars...` catalog is only for ComCamSim. (The outputName below looks correct however.)

pipelines/_ingredients/LSSTComCam/DRP.yaml
```diff
@@ -75,7 +75,8 @@ tasks:
     class: lsst.analysis.tools.tasks.refCatSourceAnalysis.RefCatSourceAnalysisTask
     config:
       # Only run metrics for analyzing the preSources:
-      connections.data: preSourceTable_visit_gaia_dr3_20230707_match_astrom
+      connections.data: preSourceTable_visit_uw_stars_20240524_match_astrom
```
cmsaunders, line 78: Same as LSSTCam comment, what was here before should be correct.

pipelines/_ingredients/LSSTCam/DRP.yaml
```diff
@@ -108,6 +109,11 @@ tasks:
       python: |
         from lsst.analysis.tools.atools import *
   # These don't actually work now:
+  photometricMatchVisit:
+    class: lsst.analysis.tools.tasks.photometricCatalogMatch.PhotometricCatalogMatchVisitTask
+    config:
+      extraColumns: ["x", "y", "ap09Flux", "ap09FluxErr"]
```
cmsaunders, line 115: Do these have to go here because they're not present when `photometricMatchPreVisit` is run and so would cause that task to fail if they were in `obs_subaru`?

pipelines/_ingredients/LSSTCam/DRP.yaml
```diff
@@ -108,6 +109,11 @@ tasks:
...
+      referenceCatalogLoader.doApplyColorTerms: false
```
cmsaunders, line 116: Can you make this change in `obs_lsst` like you do for `photometricCatalogMatchTask`?

pipelines/_ingredients/LSSTCam/DRP.yaml
```diff
@@ -116,6 +122,7 @@ tasks:
       connections.catalog: preSourceTable_visit
       connections.targetCatalog: preSourceTable_visit
       connections.matchedCatalog: preSourceTable_visit_gaia_dr3_20230707_photoMatch
+      referenceCatalogLoader.doApplyColorTerms: false
```
cmsaunders, line 125: Same as above, can't you set this in `obs_lsst`?

pipelines/_ingredients/LSSTComCam/DRP.yaml
```diff
@@ -108,6 +109,11 @@ tasks:
       python: |
         from lsst.analysis.tools.atools import *
   # These don't actually work now:
+  photometricMatchVisit:
+    class: lsst.analysis.tools.tasks.photometricCatalogMatch.PhotometricCatalogMatchVisitTask
+    config:
+      extraColumns: ["x", "y", "ap09Flux", "ap09FluxErr"]
+      referenceCatalogLoader.doApplyColorTerms: false
```
cmsaunders, line 116: Same as above, can you set this in `obs_lsst`?

cmsaunders (commented):
This mostly looks good. I think a few things could be moved to `obs_lsst`. The only thing that definitely needs to be changed is the `astrometricRefCatPreSourceVisit` input name. Let me know if I can help with figuring out the error that led to that change.

pipelines/_ingredients/LSSTCam/DRP.yaml
leeskelvin: I think this was changed in an effort to diagnose the cascade of errors. Changed back, here and for LSSTComCam.

pipelines/_ingredients/LSSTComCam/DRP.yaml
leeskelvin: Changed back to `preSourceTable_visit_gaia_dr3_20230707_match_astrom`.

pipelines/_ingredients/LSSTCam/DRP.yaml
leeskelvin: If you're referring to the `extraColumns`, I have no opinion on those. They were simply copied in verbatim from `$ANALYSIS_TOOLS_DIR/pipelines/visitQualityCore.yaml`, a YAML which gets imported into this YAML. That makes `photometricMatchVisit` visible in this YAML. I needed to over-ride the `referenceCatalogLoader.doApplyColorTerms` config to `false`, which necessitates me redefining the entire task and prior configs again here.

leeskelvin:
Thanks @cmsaunders, configs have been moved to `obs_lsst` where appropriate, significantly cleaning up this PR and removing 1 commit.

pipelines/_ingredients/LSSTCam/DRP.yaml
cmsaunders: I thought this was the case, but in a recent slack thread there was a big discussion about how the configs are actually supposed to be inherited.

cmsaunders (approved):

PR #156, opened 2024-09-13
DM-46342: Do not write artifact masks

arunkannawadi:


arunkannawadi:
Closing without merging, as this was a quick hack to get DC2 reprocessing with w37 underway. Actual ticket and fix is in `drp_tasks`.

PR #157, opened 2024-09-17
DM-43077: Convert all tasks to use CalibrateImageTask outputs

parejkoj:


pipelines/LSSTComCamSim/quickLook.yaml
```diff
@@ -12,45 +12,15 @@ tasks:
       doBrighterFatter: false
   # TO DO: Replace with calibrateImage in a way that we can use the
   # preSources downstream.
-  characterizeImage:
-    class: lsst.pipe.tasks.characterizeImage.CharacterizeImageTask
-    config:
-      doApCorr: false
-      doDeblend: false
```
arunkannawadi, line 19: Could you comment briefly on Jira about the lack of this option in `calibrateImage`? deblending seems to be always on but they have been turned off in `characterizeImage`.

arunkannawadi (changes requested):

pipelines/LSSTComCamSim/quickLook.yaml
parejkoj: Deblending in `calibrateImage` is generally quite fast, because the detection threshold is high, and only happens after psf sources are found, not before. In `characterizeImage` it wasn't strictly necessary (see DM-36958) since that task was only finding psf sources and `objectSizeStarSelector` doesn't care about blendedness. There's no "turn off this step" options in `calibrateImage`, because all of the steps are either necessary for the output, or so fast you don't notice them.

arunkannawadi: I was less concerned about the speed but a [comment](https://github.com/lsst/pipe_tasks/blob/1d79ad6392c0d10a15cf81645d5fb2448bb5c88a/python/lsst/pipe/tasks/characterizeImage.py#L233-L234) in the code said turning it on messes things up. That comment is 8 years old now, but DM-36958 seem to have gotten stuck in want of an RFC. That's what has left me confused here.

arunkannawadi (approved):

pipelines/LSSTComCamSim/quickLook.yaml
parejkoj: The important thing here is that we never want to deblend during psf determination, and calibrateImage doesn't.

PR #162, opened 2024-10-08
DM-46708: switch back to MakeWarpTask

TallJimbo:


bps/resources/LSSTCam-imSim/DRP-DP0.2.yaml
```diff
@@ -10,10 +10,6 @@
 #
 pipetask:
   # DRP tasks
-  makeDirectWarp:
-    requestMemory: 8192
-  makePsfMatchedWarp:
-    requestMemory: 8192
```
kherner, line 16: I don't see a corresponding addition of makeWarp here. Does it fit it in the default memory request?

pipelines/_ingredients/DECam/DRP.yaml
```diff
@@ -46,24 +46,21 @@ tasks:
...
       connections.skyWcsName: "jointcal"
-  makeDirectWarp:
-    class: lsst.pipe.tasks.make_direct_warp.MakeDirectWarpTask
+  makeWarp:
+    class: lsst.pipe.tasks.makeWarp.MakeWarpTask
     config:
       # These configs match obs_subaru, to facilitate 1:1 comparisons between DECam and HSC
-      warper.warpingKernelName: lanczos5
+      matchingKernelSize: 29
+      modelPsf.defaultFwhm: 7.7
```
kherner, line 54: I am not sure what the correct values are here, so I will only suggest double-checking them.

kherner (approved):
Mostly looks good, but see the comments, especially on the bps yaml.

bps/resources/LSSTCam-imSim/DRP-DP0.2.yaml
TallJimbo: The history here is that prior to the ticket that I'm mostly reverting, there was only a configuration of `makeWarp` with these values, and that ticket replaced it with configuration for `makeDirectWarp` and `makePsfMatchedWarp` (instead of adding those).  But I think there didn't need to be an override for any of these, because they really should fit in the default memory request, and that's consistent with the lack of memory overrides in `test-med-1.yaml` here.  I'm pretty sure there was just a one-off problem in the DP0.2 era that has long since been fixed.

PR #165, opened 2024-10-14
DM-46850: Replace makeWarp for the second time

arunkannawadi:


pipelines/LSSTCam/DRP.yaml
TallJimbo, line 1: This file was intentionally deleted; I think restoring it here is a rebase accident.

(The idea is that we'll restore it in the future from the LSSTComCam one, but in the meantime we don't want to maintain it.)

pipelines/LSSTCam/nightly-validation.yaml
TallJimbo, line 1: Restoring this file is another rebase accident.

TallJimbo (approved):

PR #169, opened 2024-10-23
DM-47068: Remove faro from pipelines.

TallJimbo:


bps/resources/LSSTCam-imSim/DRP-test-med-1.yaml
```diff
@@ -1,5 +1,5 @@
 # This BPS configuration snippet adds updated requestMemory values for all DRP
-# and faro tasks known to frequently exceed the default in regular ImSim-DC2
+# asks known to frequently exceed the default in regular ImSim-DC2
```
leeskelvin, line 2: tasks typo

pipelines/HSC/DRP-RC2_subset.yaml
```diff
@@ -405,20 +181,6 @@ subsets:
       'tract' or 'patch' as part of the data ID expression if all
       reference catalogs or diffIm templates that cover these
       detector-level quanta are desired.
-  nightlyStep5:
```
leeskelvin, line 408: I think this change necessitates that the relevant lines in the `rc2_subset` repo are also removed:
https://github.com/lsst/rc2_subset/blob/1a133e3ce3b2ff950d60b48526d126463bd35fc6/bin/run_rc2_subset.sh#L119. If this isn't removed, then I'd expect any nightly test to fail when it gets to `nightlyStep5`.

leeskelvin: There's also mention of `nightlyStep5` inside the legacy `measureHscRC2Metrics.sh` script, which also calls `faro` at points. This script has been superceded by `run_rc2_subset.sh`, and I wouldn't object to it simply being removed at this point.

leeskelvin (approved):
I think this is mostly fine, but the removal of `nightlyStep5` in `rc2_subset` necessitates that this step is also removed from the `run_rc2_subset.sh` script inside the `rc2_subset` repo.

pipelines/HSC/DRP-RC2_subset.yaml
leeskelvin: Edit: please disregard the above. I clicked your PR links on the Jira ticket, but one them was duplicated and `rc2_subset` was missing, so I didn't realize there was a PR for `rc2_subset` as well!

I still wouldn't object if you'd rather just remove the entire `measureHscRC2Metrics.sh` script however, as it should probably not be used any longer.

PR #171, opened 2024-10-24
DM-46188: fix and test LSST[Com]Cam[Sim] pipelines with the step-tester

TallJimbo:
This doesn't guarantee that these pipelines would actually run, but it at least ensures they stay self-consistent.

### DIFF
--- a/tests/test_pipelines.py
+++ b/tests/test_pipelines.py
@@ -687,13 +687,19 @@ class PipelineTestCase(unittest.TestCase):
             os.path.join(PIPELINES_DIR, "LSSTComCam", "DRP.yaml"),
             [
                 "#step1",
-                # TODO[DM-47010]: all other steps are broken
+                "#step2a",
+                "#step2b",
+                "#step2d",
+                "#step2e",
+                "#step3",
+                "#step4",
+                "#step5",
+                "#step6",
+                "#step7",
             ],
             initial_dataset_types=REFCATS,
             expected_inputs=COMMON_INPUTS | LSSTCOMCAM_INPUTS,
-            # TODO[DM-47010]: once we run the rest of the steps this should
-            # include at least COMMON_OUTPUTS.
-            expected_outputs={"preSourceTable", "calexp"},
+            expected_outputs=COMMON_OUTPUTS,
         )
         tester.run(butler, self)
 
--- a/tests/test_pipelines.py
+++ b/tests/test_pipelines.py
@@ -99,7 +99,6 @@ LSSTCOMCAM_INPUTS = {
     "linearizer",
     "defects",
     "the_monster_20240904",
-    "gaia_dr3_20230707",
 }
 
 # LSSTComCamSim common inputs, in addition to COMMON_INPUTS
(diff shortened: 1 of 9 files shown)


tests/test_pipelines.py
```diff
@@ -82,6 +82,32 @@
     "truth_summary",
 }
 
+# LSSTCam common inputs, in addition to COMMON_INPUTS
+LSSTCAM_INPUTS = {
+    "bfk",
+    "linearizer",
+    "defects",
+    "the_monster_20240904",
+    "gaia_dr3_20230707",
```
erykoff, line 91: I think that the astrometric refcat should be the monster as well.

TallJimbo: I wasn't sure if it was time to make that switch but I'm happy to do it.

erykoff: I just merged the fix to meas_algorithms so it's good to go.  Also we don't have gaia_dr3 on the summit.

tests/test_pipelines.py
```diff
@@ -82,6 +82,32 @@
...
+}
+
+# LSSTComCam common inputs, in addition to COMMON_INPUTS
+LSSTCOMCAM_INPUTS = {
+    "bfk",
+    "ptc",
+    "linearizer",
+    "defects",
+    "the_monster_20240904",
+    "gaia_dr3_20230707",
```
erykoff, line 101: This should be just the_monster as well.

tests/test_pipelines.py
```diff
@@ -220,6 +246,24 @@
...
+    "calexp", "visitSummary"
+}
+
+# All refcats used by any pipelines, for inclusion as the initial_dataset_types
+# argument to PipelineStepTester; the intent is just to allow the 'skypix'
+# dimension placeholder used in these connections to be resolved.
+REFCATS = [
+    ("ps1_pv3_3pi_20170110", {"htm7"}, "SimpleCatalog", False),
+    ("gaia_dr2_20200414", {"htm7"}, "SimpleCatalog", False),
+    ("gaia_dr3_20230707", {"htm7"}, "SimpleCatalog", False),
```
erykoff, line 260: I think it's okay to have this here even though we don't have all these refcats on the summit.

erykoff: Though do we need a subset that is on the summit and a fuller set that is at DFs and make sure that quicklook uses only summit refcats?

erykoff (approved):
Looks good, but some questions about summit refcats and quicklook pipelines.

tests/test_pipelines.py
TallJimbo: This is just a convenience list of all the refcats anywhere (note that it includes simulated ones).  The real check on which refcat is used in a pipeline is the variables used to supply `expected_inputs` to the tester.

erykoff: Ah!  I see.  Then my comment above that gaia_dr3 should be removed from the lsst inputs.

erykoff (approved):

PR #172, opened 2024-10-25
DM-46627: Add MultiProFit tasks to DRP pipelines

taranu:


pipelines/HSC/DRP-ci_hsc.yaml
```diff
@@ -75,6 +75,14 @@ tasks:
...
       # and parallax catalog available to use here.
       applyAstrometricCorrections: false
+  fitDeblendedObjectsSersicMultiprofit:
+    class: lsst.meas.extensions.multiprofit.pipetasks_fit.MultiProFitCoaddSersicFitTask
+    config:
+      fit_coadd_multiband.bands_fit: ["r", "i"]
+  analyzeMultiprofitCore:
+    class: lsst.analysis.tools.tasks.ObjectTableTractAnalysisTask
+    config:
+      bands: ["r", "i"]
```
fred3m, line 85: How much time does this add to CI packages? Will the fact that we only deblend 10 sources in scarlet affect this at all (since most parents were not deblended)? I'm guessing it's fine, just thought that I should point it out in case you hadn't thought about it.

fred3m (commented):
LGTM, just a comment to make sure that you've thought about the effects of the deblender only running on 10 blends in CI.

pipelines/HSC/DRP-ci_hsc.yaml
taranu: On sdfiana013 the 6 PSF fitting tasks each took 75-95s while the multiband Sersic fit took 172s, so it'll be about 5 mins if using `-j 6` or more (the extra matching and plot tasks take another 30s or so but will be irrelevant on non-CI datasets).

PR #173, opened 2024-10-30
DM-47263: Simplify amp offset correction logging and config

leeskelvin:


erykoff (approved):

pipelines/_ingredients/HSC/DRP.yaml
```diff
@@ -14,7 +14,7 @@ tasks:
     class: lsst.ip.isr.IsrTask
     config:
       doAmpOffset: true
-      ampOffset.doApplyAmpOffset: true
+      ampOffset.measureOnly: false
```
erykoff, line 17: Is this necessary since this is the task default?

leeskelvin: Yes - Jenkins was failing otherwise. The reason: `pipelines/_ingredients/HSC/DRP.yaml` imports `pipelines/_ingredients/DRP-full.yaml`, which over-rides `ampOffset.measureOnly` to `True`. Therefore, it's necessary to revert the `measureOnly` for HSC DRP processing in order to apply it as well.

leeskelvin: I think the problem here is that we overrode this config at all in `pipelines/_ingredients/DRP-full.yaml`. I'm not probably of the mind to go the other way, and remove the config over-ride from DRP-full entirely, and then only over-ride this in non- `_ingredients` pipelines as we need this functionality.

leeskelvin:
Withdrawing this PR, due to pushback on proposed config changes.

PR #174, opened 2024-10-30
DM-47210: Add analysis tasks to the injected RC2 pipelines.

jtmccann:


taranu (approved):

pipelines/LSSTComCam/DRP-post-injected.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+      bands: ["u", "g", "r", "i", "z", "y"]
+      python: |
+        from lsst.analysis.tools.atools.diffMatched import reconfigure_diff_matched_defaults
+
+        reconfigure_diff_matched_defaults(
+          config=config,
+          context="injection",
+          key_flux_meas="psf",
+          bands_color = {"u", "g", "r", "i", "z", "y"},
```
taranu, line 43: This should be okay to omit since ugrizy is the default.

pipelines/LSSTComCam/DRP-post-injected.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+      # TODO: Remove as part of DM-44139
+      columns_ref_mag_to_nJy: {u_mag: "u_flux", g_mag: "g_flux", r_mag: "r_flux", i_mag: "i_flux", z_mag: "z_flux", y_mag: "y_flux"}
+      columns_ref_copy: [
+        "injected_id", "source_type",
+        "u_mag", "g_mag", "r_mag", "i_mag", "z_mag", "y_mag",
+        "injection_flag", "injected_isPatchInner",
+      ]
+      columns_target_coord_err: ["coord_raErr", "coord_decErr"]
+      columns_target_copy: [
```
taranu, line 22: I think I forgot `objectId` here.

SConstruct
```diff
@@ -18,7 +19,25 @@ scripts.BasicSConstruct(
 )
 PKG_ROOT = env.ProductDir("drp_pipe")
 
-# make deepCoadd injection pipelines for rc2_subset and RC2
+additional_pipeline_RC2 = os.path.join(
+    PKG_ROOT,
+    "pipelines",
+    "HSC",
+    "DRP-RC2-post-injected.yaml",
```
taranu, line 26: Can we append `-brightstars` to this? Or `bright-stars`? Whatever we want to call this (@leeskelvin ?).

SConstruct
```diff
@@ -18,7 +19,25 @@ scripts.BasicSConstruct(
...
+)
+
+additional_pipeline_LSSTComCam = os.path.join(
+    PKG_ROOT,
+    "pipelines",
+    "LSSTComCam",
+    "DRP-post-injected.yaml",
```
taranu, line 33: Same here, add `-brightstars`.

pipelines/HSC/DRP-RC2-post-injected.yaml
```diff
@@ -0,0 +1,46 @@
+description: Match tract-level injected_deepCoadd_catalog datasets to injected_objectTable_tract datasets.
+instrument: lsst.obs.subaru.HyperSuprimeCam
+imports:
+  - location: $SOURCE_INJECTION_DIR/pipelines/match_injected_tract_catalog.yaml
+  - location: $ANALYSIS_TOOLS_DIR/pipelines/coaddDiffMatchedInjectedCore.yaml
+tasks:
+  compare_object_to_injected:
```
taranu, line 7: `compare_object_to_injected_bright_stars`?

pipelines/HSC/DRP-RC2-post-injected.yaml
```diff
@@ -0,0 +1,46 @@
...
+      ]
+      columns_target_coord_err: ["coord_raErr", "coord_decErr"]
+      columns_target_copy: [
+        "g_psfFlux", "r_psfFlux", "i_psfFlux", "z_psfFlux", "y_psfFlux",
+        "g_psfFluxErr", "r_psfFluxErr", "i_psfFluxErr", "z_psfFluxErr", "y_psfFluxErr",
+        "g_psfFlux_flag", "r_psfFlux_flag", "i_psfFlux_flag", "z_psfFlux_flag", "y_psfFlux_flag",
+        "patch", "detect_isDeblendedSource", "detect_isPatchInner", "detect_isPrimary", "merge_peak_sky",
+        "refExtendedness", "refSizeExtendedness",
+      ]
+  diff_matched_analysis:
```
taranu, line 29: I think it's time to give this a better name too - `analyze_matched_injected_bright_stars`?

SConstruct
```diff
@@ -18,7 +19,25 @@ scripts.BasicSConstruct(
...
+)
+
+subset_name = "injected_coadd_analysis"
```
taranu, line 36: Maybe `injected_bright_stars_coadd_analysis`?

taranu (changes requested):
In general I think this looks good (if it works, it works!), and I'm okay to have this go in before the DC2-based injection, but I think there are some things we'll want to pre-emptively specify stars or bright stars in the names for so they don't clash with future injection pipelines/tasks. I'm pinging @leeskelvin for thoughts on that.

SConstruct
jtmccann: Could these post-injected pipelines be generalized to be used for any kind of injection? Rather than making multiple different post-injected pipelines with slightly different configs. For example, if we wanted to do galaxy injections what additional columns would we need? If it's just a few then it might be easier just to include them by default for generality's sake.

taranu: Hmm... not really. See the [equivalent file I made](https://github.com/lsst-sitcom/sciunit_galaxies/blob/a7316be59e4a694dc019e9607e97ff4e3976d904/pipelines/DRP-post-DC2-injected.yaml) for galaxy injection. We can harmonize the copied columns without doing any harm (besides bloating the matched tables, but it's conceivable we still might want to look at model/aperture mags for stars) but there are more config changes than just that.  The consolidate task needs to have added columns for galaxy parameters that won't exist in injection catalogs with just stars. The analysis task could make the same plots/metrics but if you're only injecting stars then the ones for reference galaxies will be empty/useless and the ones for reference objects of all types will be identical to the ones for stars alone.

leeskelvin: I can certainly see the merit for somehow distinguishing between the bright-star flavor and the galaxy flavor of these injection YAMLs. I'm not sure if `bright-stars` is sufficient long-term, as we may have another "bright stars" catalog that we want to inject down the line. Perhaps we need to embed the `DM-NNNNN` name that corresponds to the recommended input injection catalog?

Or, rather than "bright stars", if this injection YAML is instead suitable for all "stars only" injection catalogs, then lets just append `-stars`? That would mean that the ultimate final user-facing YAML would be named, e.g., `DRP-RC2_subset+injected_deepCoadd-stars.yaml`.

Naming things is hard (TM), so for now, I propose that we append `-stars` to this catalog and `-galaxies` to the future galaxies catalog. Longer-term, we may have to loop back to derive a better and more concise naming scheme.

taranu: I'm fine with `-stars`. The pipeline itself should presumably work with any kind of injection catalog with stars in it.

`galaxies` is also fine for now, I suppose. The pipeline would need to change if it's only galaxies vs galaxies + stars. For now we can document that users should include stars too - galaxy photometry algorithms don't necessarily need to perform very well on stars but they should at least not fail and ideally not be much worse than PSF mags for anyone doing a homogeneous selection/running classifiers.

If/when we get around to injecting non-parametric galaxies (or a mix of both) we might need to tweak things further.

pipelines/HSC/DRP-RC2-post-injected.yaml
jtmccann: I tried changing the task names and got an error about the same dataset type being the output of two different tasks. I decided to revert the task name change since there are only config differences between a star vs galaxy run. I don't think a star run is fundamentally different from a galaxy run, so they ought to keep the same dataset type and task names.

I will however keep the "-post-injected-stars" names for the YAMLs to differentiate the star configs from any future galaxy config files.

Let me know what you think. Thanks!

taranu (approved):

PR #183, opened 2024-11-22
DM-46991: switch to calibrateImage

TallJimbo:


pipelines/LSSTComCamSim/nightly-validation.yaml
```diff
@@ -15,15 +15,8 @@ tasks:
...
       connections.outputName: preSourceTableCore
-  calibrate:
-    class: lsst.pipe.tasks.calibrate.CalibrateTask
-    config:
-      python: |
-        from lsst.analysis.tools.tasks import CalexpSummaryAnalysisTask
-        config.createSummaryMetrics.retarget(CalexpSummaryAnalysisTask)
-        from lsst.analysis.tools.atools import CalexpSummaryMetrics
-        config.createSummaryMetrics.atools.calexpSummaryMetrics = CalexpSummaryMetrics
-      doCreateSummaryMetrics: true
```
natelust, line 26: What did you do about getting these metrics out?

pipelines/_ingredients/DRP-minimal-calibration.yaml
```diff
@@ -10,10 +10,61 @@ imports:
...
+        # Table schema standardization currently works by looking at these
+        # 'local' calibration columns and using them to re-calibrate the raw
+        # measurements, despite the fact that the calibrateImage outputs also
+        # include already-calibrated columns.
+        config.star_measurement.plugins.names |= ["base_LocalPhotoCalib", "base_LocalWcs"]
+
+  transformSourceTable:
+    class: lsst.pipe.tasks.postprocess.TransformSourceTableTask
+    config:
+      functorFile: $PIPE_TASKS_DIR/schemas/PreSource.yaml
```
natelust, line 67: I think this is the second time I have seen this, it seems to be pointing to pipe_tasks anyway, Is there a reason it is not the default?

pipelines/_ingredients/LSSTCam-imSim/DRP.yaml
```diff
@@ -34,14 +37,8 @@ tasks:
...
-  calibrate:
-    class: lsst.pipe.tasks.calibrate.CalibrateTask
-    config:
-      connections.astromRefCat: "cal_ref_cat_2_2"
-      connections.photoRefCat: "cal_ref_cat_2_2"
-      python: |
-        config.astromRefObjLoader.filterMap = parameters.filterMap
-        config.photoRefObjLoader.filterMap = parameters.filterMap
+  calibrateImage:
+    class: lsst.pipe.tasks.calibrateImage.CalibrateImageTask
```
natelust, line 41: Those connections can safely change?

pipelines/_ingredients/LSSTCam/DRP.yaml
```diff
@@ -7,8 +7,11 @@ imports:
   - $ANALYSIS_TOOLS_DIR/pipelines/coaddQualityExtended.yaml
   - $ANALYSIS_TOOLS_DIR/pipelines/diaTractQualityCore.yaml
   - $ANALYSIS_TOOLS_DIR/pipelines/matchedVisitQualityCore.yaml
-  - $ANALYSIS_TOOLS_DIR/pipelines/visitQualityCore.yaml
+  - location: $ANALYSIS_TOOLS_DIR/pipelines/visitQualityCore.yaml
+    exclude:
+      - analyzeCalibrateMetadata
```
natelust, line 12: If you keep excluding this everywhere, it seems to tell me that maybe this does not believe in the visitQualityCore, as that is defined explicitly as the tasks we want to always run

pipelines/_ingredients/LSSTComCam/DRP.yaml
```diff
@@ -11,6 +13,16 @@ imports:
   - $ANALYSIS_TOOLS_DIR/pipelines/wholeSkyCore.yaml
 tasks:
   isr: lsst.ip.isr.IsrTaskLSST
+  characterizeImage: lsst.pipe.tasks.characterizeImage.CharacterizeImageTask
```
natelust, line 16: make a note here why you are not using the thing someone would see everywhere else (calibrateImage)

pipelines/LSSTCam/quickLook.yaml
```diff
@@ -22,47 +22,14 @@ tasks:
...
-          config.measurement.slots.shape = "base_SdssShape"
-          config.measurement.slots.psfShape = "base_SdssShape_psf"
-          config.measurement.plugins['base_PixelFlags'].masksFpAnywhere.remove('STREAK')
-          config.measurement.plugins['base_PixelFlags'].masksFpCenter.remove('STREAK')
-          config.doApCorr = False
-          config.doDeblend = False
-          config.astrometry.sourceSelector["science"].doRequirePrimary = False
-          config.astrometry.sourceSelector["science"].doIsolated = False
+  calibrateImage:
+    class: lsst.pipe.tasks.calibrateImage.CalibrateImageTask
```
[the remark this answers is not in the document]
TallJimbo: I'll have to look into this into more detail on [DM-47632](https://rubinobs.atlassian.net/browse/DM-47632), where I'll switch ComCam `quickLook` (the only really important variant of `quickLook`) over to `calibrateImage`.  For now:
 - I'm not sure we should even have a version of LSSTCam `quickLook` that tries to do any on-sky processing, because it'll surely get out of date before it becomes useful, but I assume Merlin added it because it was useful for something.  I hope the [DM-47320](https://rubinobs.atlassian.net/browse/DM-47320) refactor will land first and basically reset it to a reference to a base pipeline shared with the ComCam one.
 - `calibateImage` and the `quickLook` version of `characterizeImage` and `calibrate` _have_ been optimized (independently) for very similar goals - fast, no frills processing, in Prompt Processing and Rapid Analysis, respectively.  So it's at least plausible there will not be an need to change the default configurations.

pipelines/LSSTComCamSim/nightly-validation.yaml
TallJimbo: There's a separate `PipelineTask` (label `calexpSummary`) that reads the calexp summaries and writes the metrics that was already being used instead in more actively-maintained pipelines, and I've switched this one to behave the same way.

pipelines/LSSTComCamSim/quickLook.yaml
```diff
@@ -12,45 +12,15 @@ tasks:
...
-          config.measurement.slots.shape = "base_SdssShape"
-          config.measurement.slots.psfShape = "base_SdssShape_psf"
-          config.measurement.plugins['base_PixelFlags'].masksFpAnywhere.remove('STREAK')
-          config.measurement.plugins['base_PixelFlags'].masksFpCenter.remove('STREAK')
-          config.doApCorr = False
-          config.doDeblend = False
-          config.astrometry.sourceSelector["science"].doRequirePrimary = False
-          config.astrometry.sourceSelector["science"].doIsolated = False
+  calibrateImage:
+    class: lsst.pipe.tasks.calibrateImage.CalibrateImageTask
```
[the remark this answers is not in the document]
TallJimbo: This has sort of the opposite problem as `LSSTCam/quickLook`, in that I think it's basically only run in the past instead only being run in the future. But the answer is the same.

pipelines/_ingredients/DRP-minimal-calibration.yaml
TallJimbo: It's a mess:
- There is one configuration of this task in the "minimal-calibration" pipeline (used only by ImSim, despite the fact that many other pipelines don't run FGCM or GBDES), which consumes the outputs of `calibrateImage` and hence has to use `PreSource.yaml` to work.  But it's called `transformSourceTable`, not `transformPreSourceTable`.
- There are two configurations of this task in the "full" pipeline.  But the `transformPreSourceTable` is the one that's configured like the "minimal-calibration" pipeline's `transformSourceTable`, while the "full" version of `transformSourceTable` uses the default `Source.yaml` because it runs on the output of `writeRecalibratedSourceTable`.

My plan to fix it has two prongs:
 - DM-34875 will make `writeRecalibratedSourceTable` produce an `ArrowAstropy` table, just like `calibrateImage`, and then we can delete `PreSource.yaml` entirely.
 - DM-47320 will make ImSim behave like the other pipelines that don't use FGCM or GBDES, and use task labels that stay more consistent with the configuration.

pipelines/_ingredients/LSSTCam-imSim/DRP.yaml
TallJimbo: Yes. It's unfortunate that we're changing refcats here in the pipelines at all, cause in most other cases we do it in obs_ configs and we recently had some agreement on Slack that obs_ configs were preferred.  But I didn't want that scope on this ticket.

pipelines/_ingredients/LSSTCam/DRP.yaml
TallJimbo: I'll delete it on [DM-47632](https://rubinobs.atlassian.net/browse/DM-47632), and in fact I had originally deleted it on an earlier version of this ticket before I realized I needed to keep ComCam using the old tasks for a bit longer.  I just need to keep it as long as any instrument is still using `calibrate`.

PR #195, opened 2024-12-18
DM-47940: Remove analysis_drp.

TallJimbo:


erykoff (approved):
This all looks good.  The yaml checker also noticed that in `DRP-minimal-calibration.yaml` that there is a `True` that should be `true` which maybe should be fixed here while we're making changes?

PR #198, opened 2025-01-07
DM-48145: Switch back to initial visitSummary in preSource analysis.

TallJimbo:
This counteracts a defaults change upstream in analysis_tools.


leeskelvin (approved):
LGTM.

PR #207, opened 2025-02-06
DM-47738: Add deconvolution to pipelines where appropriate

fred3m:


taranu (approved):

tests/tmp0qrt90_l/butler.yaml
```diff
@@ -0,0 +1,14 @@
+datastore:
```
taranu, line 1: Uh... I think you accidentally committed this temp file and the sqlite db instead of deleting them.

But while you're at it, you may as well add `tests/tmp*` to the root `.gitignore`.

taranu (commented):
See comments.

tests/tmp0qrt90_l/butler.yaml
fred3m: Hmmm, I didn't add those files. But I do remember one time that I was running a test and my connection to USDF broke. Maybe that left some straggler files, so adding that directory to `.gitignore` sounds like a good idea.

tests/test_drp-v2.py
```diff
@@ -233,7 +233,7 @@ def check_stage(
         step_member_union = set()
         for step_label in step_labels:
             step_member_union.update(pipeline_graph.task_subsets[step_label])
-        self.assertEqual(step_member_union, stage_members)
+        self.assertSetEqual(step_member_union, set(stage_members))
```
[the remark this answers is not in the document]
taranu: I understand the use of `assertSetEqual`, it's the cast of `stage_members` that I'm surprised by. If it wasn't a set before then how did this pass? What changed to make the cast necessary?

PR #209, opened 2025-02-11
DM-48591: Add multiprofit tasks to all pipelines

taranu:


pipelines/_ingredients/LSSTCam-imSim/DRP.yaml
```diff
@@ -76,19 +76,22 @@ tasks:
       python: |
         from lsst.pipe.tasks.diff_matched_tract_catalog import MatchedCatalogFluxesConfig
         columns_flux = {}
-        fluxes_meas_psf = []
+        fluxes_meas_other = []
+        models_other = ("psf", "gaapOptimal", "kron", "sersic")
```
TallJimbo, line 80: My understanding is that `gaapOptimal` is actually the least useful of the GaAP fluxes (the additional DOF needed to determine the aperture outweighs typically the fact that the aperture is a better match), and I'd like to remove it from the defaults if I remember and find the time.  Is it included here because it's the most total-flux-like of them?

taranu: I was hoping it had improved since being introduced ~3 years ago but in practice it seems to be really bad, so I supposed I'll replace it with whatever is deemed to be best (old messages suggest 1p0).

PR #212, opened 2025-02-18
DM_48896: Add yaml for DRP processing with sysnthetic source injection for diffim

BrunoSanchez:
Add a pipeline yaml that takes DRP Diffim stage and adds injected synthetic sources, for transient discovery testing.


pipelines/LSSTComCam/DRP_InjectedPVIDiffim.yaml
```diff
@@ -0,0 +1,345 @@
+description: DRP Pipeline for LSSTComCam with injected point sources
+instrument: lsst.obs.lsst.LsstComCam
+
+imports:
+  - location: $DRP_PIPE_DIR/pipelines/LSSTComCam/DRP.yaml
+
+parameters:
+  fakesType: 'injected_'
+  injection_prefix: 'injection_'
+  injected_prefix: 'injected_'
```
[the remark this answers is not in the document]
leeskelvin: Clear for what purpose though? If we change the task default from `injected_` to `post_injection_` for example, the `injected_prefix` here would still read `injected_`. Someone might look at this thinking your injected outputs will still be `injected_`.

pipelines/LSSTComCam/DRP_InjectedPVIDiffim.yaml
```diff
@@ -0,0 +1,345 @@
...
+  inject_visit:
+    class: lsst.source.injection.inject_visit.VisitInjectTask
+    config:
+      external_psf: false
+      external_photo_calib: false
+      external_wcs: false
+      connections.input_exposure: pvi
+      connections.output_exposure: injected_pvi
+      connections.output_catalog: injected_pvi_catalog
+      selection: "np.isin(injection_catalog['visit'], {visit})"
```
leeskelvin, line 22: Can I ask why you've decided to hard-code the `VisitInjectTask` here, rather than dynamically create it as we do with other injection pipelines? Putting this in would be the second instance in the stack (after `pipelines/_ingredients/ApPipeWithFakes.yaml`).

pipelines/LSSTComCam/DRP_InjectedPVIDiffim.yaml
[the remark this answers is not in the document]
BrunoSanchez: That's fine then. Since this yaml defines inputs and outputs explicitly and no dataset type templates are _actually_ in use, I will remove these.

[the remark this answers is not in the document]
leeskelvin: The alternative is that you do use these parameters in your config over-rides below.

pipelines/LSSTComCam/DRP_InjectedPVIDiffim.yaml
```diff
@@ -0,0 +1,345 @@
...
+  injectedMatch:
+    class: lsst.ap.pipe.matchSourceInjected.MatchPVIInjectedTask
```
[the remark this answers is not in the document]
leeskelvin: Perhaps not, as making that switch returns:
```
AttributeError: lsst.pipe.base.config.MatchInitialPVIInjectedConfigConnections has no attribute injectedPVICat
```
I'm on `main` in `drp_pipe` and `ap_pipe` - do I need to set up something else?

pipelines/LSSTComCam/DRP_InjectedPVIDiffim.yaml
BrunoSanchez: Because dynamically creating this has a key shortcomings for our case, as it doesn't include the matching tasks, with the associated metrics tasks to verify that we have found our injections

pipelines/LSSTComCam/DRP_InjectedPVIDiffim.yaml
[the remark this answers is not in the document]
BrunoSanchez: I made a mistake, and I was using a branch class from `ap_pipe`. I think I can update this to use the current main function, as a replacement matching task is still in flux.

pipelines/LSSTComCam/DRP_InjectedPVIDiffim.yaml
leeskelvin: From `make_injection_pipeline --help`:
```
  -a, --additional-pipelines FILE [FILE ...]
      Location(s) of additional input pipeline definition YAML file(s).Tasks
      from these additional pipelines will be added to the output injection
      pipeline.
```
This allows you to define the additional task stubs you need in their own stand-along YAML, and then build a complete SI YAML when `drp_pipe` gets constructed.

pipelines/LSSTComCam/DRP_InjectedPVIDiffim.yaml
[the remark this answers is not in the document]
BrunoSanchez: It should be fixed on new version.

ups/drp_pipe.table
```diff
@@ -14,6 +14,7 @@ setupRequired(jointcal)
 setupRequired(fgcmcal)
 setupRequired(meas_transiNet)
 setupRequired(cp_verify)
+setupRequired(ap_pipe)
```
timj, line 17: @TallJimbo I think this is the first time that a _pipe package depends on another _pipe package so I think you may want to jump in and declare that this is fine or not.

BrunoSanchez: Migrated the relevant tasks to `pipe_tasks`.

SConstruct
```diff
@@ -3,20 +3,12 @@ import os
 
 from lsst.sconsUtils import scripts
 from lsst.sconsUtils.state import env
+from lsst.sconsUtils import targets
```
leeskelvin, line 6: Merged into the above
```
from lsst.sconsUtils import scripts, targets
```
I wouldn't object if you ran `isort` and `black`/`ruff` on this entire script either; there are quite a lot of formatting errors.

SConstruct
```diff
@@ -97,9 +89,76 @@ LSSTComCam_injected = [
     )
     for suffix, pipeline_post in injected_pipelines_LSSTComCam.items()
 ]
+
+# diffim injection pipeline creation
+subset_names = "injected_diffim_analysis,injected_stage4-measure-variability"
+subset_description = "Analysis tasks for source injection for diffim catalogs"
```
leeskelvin, line 95: This isn't currently being used. That said, considering the PR comment in the `source_injection` repo, this may be switched up substantially following an API change.

leeskelvin (approved):
LGTM.

PR #214, opened 2025-02-24
DM-47320: add ComCam/DRP-v2 with new names and stages/steps concept

TallJimbo:


config/calibrateImage.py
```diff
@@ -0,0 +1,36 @@
+# DRP CONFIGURATION FOR CalibrateImageTask
+#
+# This configuration can't go in the task defaults, because those are shared
+# with AP.  It could be moved directly into a pipeline YAML after the v1->v2
+# migration is done.
+
+# In DRP not calibrating the pixels (or the backgrounds) until we have
+# our final photometric calibration makes everything downstream much
+# simpler.
+do_calibrate_pixels: False
```
leeskelvin, line 10: Is there a missing `config.` here?

pipelines/_ingredients/analysis-association-initial.yaml
leeskelvin, line 1: Just noting here to also rename this file pending any name changes that come through via #dm-naming-things. It occurs to me that the descriptions here are still using `Tier1` terminology, which is at least 2 generations old, but addressing that is well beyond the scope of this ticket.

leeskelvin (approved):
This mostly looks good to me too. A few questions here and there. I haven't had time to test this locally as you have, but I'm particularly excited to see the Correspondence class on `main`. Thanks!

config/calibrateImage.py
TallJimbo: Ah, yes, good catch!

Especially since - because I used this file to avoid duplication between the new and old pipelines - this is a rare change on this ticket that would have broken more than just the new pipeline!

pipelines/_ingredients/analysis-association-initial.yaml
TallJimbo: Yup, thanks for the reminder.

I didn't realize the "Tier 1" terminology was old (just copying what I saw in analysis_tools), and it's easy enough to remove here.

python/lsst/drp/pipe/tests/correspondence.py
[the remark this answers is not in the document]
TallJimbo: I was rather hoping it'd have a very brief shelf-life, and get deleted after we get v2 versions of all of the other major pipelines.  If you think it'd be useful after that, we should probably move it to pipe_base.  I won't do that here, but I can make a ticket for it.

python/lsst/drp/pipe/tests/correspondence.py
```diff
@@ -0,0 +1,602 @@
...
+from lsst.pipe.base.pipeline_graph import PipelineGraph, NodeType, TaskNode
+
+_LOG = logging.getLogger(__name__)
+
+
+def without_automatic_connections(names: Iterable[str]) -> set[str]:
+    return {
+        name
+        for name in names
+        if not name.endswith("_config") and not name.endswith("_log") and not name.endswith("_metadata")
```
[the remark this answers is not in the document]
TallJimbo: Given how this works right now, there's no need, because `packages` isn't added by `PipelineGraph`; it's added at a later stage (I don't recall if it's in the QG, or just written during execution without being in the QG).

python/lsst/drp/pipe/tests/correspondence.py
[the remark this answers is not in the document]
leeskelvin: I'll leave that decision up to you then! Certainly having a nice way to compare two pipelines would be useful beyond the scope of this ticket I think.

PR #216, opened 2025-02-25
DM-49145: Change string in stellar locus names from "PSFP" to "PSF"

jeffcarlin:
Note that this ticket entails an additional PR on analysis_tools


arunkannawadi (approved):

arunkannawadi:
The commits needs to be squashed here, or the commit and its revert should be dropped.

PR #218, opened 2025-02-26
DM-49155: Update LSSTComCam mock inputs for illumination corrections.

erykoff:


yalsayyad:
Approved, however you want to re-open this.   I think this might be a DM first.

PR #227, opened 2025-03-09
DM-49371: Remove (from git) machine-generated pipelines.

TallJimbo:
These were only committed accidentally.


arunkannawadi (approved):
The `SConstruct` file, at least on `main`, seems to be generating only the HSC injected pipelines. Curious how the LSSTComCam pipeline got generated on `scons` then. I also see your comment on Jira that these no longer get generated, so I guess it's not necessary to update `.gitignore` to `injected_deepCoadd*.yaml`?

TallJimbo:
I'm pretty sure I did a global search-and-replace at one point while renaming some dataset types in the commit that added them, and I think it was that rather than the SCons build that made the bad pipeline files.  I'm still not sure exactly how, but I have been reproduce getting them to come back.

PR #228, opened 2025-03-10
DM-49372: work around v1/v2 conflict in 'source' dataset type definition

TallJimbo:


pipelines/_ingredients/DRP-full.yaml
```diff
@@ -58,7 +54,16 @@ tasks:
...
+      # Use v2 name instead of old v1 name ('source') to avoid clash with v2
+      # final data product name.
+      connections.outputCatalog: recalibrated_star_unstandardized
+  transformSourceTable:
+    class: lsst.pipe.tasks.postprocess.TransformSourceTableTask
+    config:
+      # Use v2 name instead of old v1 name ('source') to avoid clash with v2
+      # final data product name.
+      connections.inputCatalog: recalibrated_star_unstandardized
+      functorFile: $PIPE_TASKS_DIR/schemas/Source.yaml
```
arunkannawadi, line 66: Curious if there's a reason why this is moved further down. The ordering doesn't matter here, right?

arunkannawadi (approved):

PR #233, opened 2025-03-19
DM-49368: Remove skyCorr from LSSTComCam pipelines

leeskelvin:
It was decided that SkyCorrectionTask should not be run on LSSTComCam datasets (see DM-47176 for more details).
As such, skyCorr does not appear in any currently run LSSTComCam steps. Excluding skyCorr from LSSTComCam pipeline YAMLs cleans up any potential contract failures that would exist as a result of this unused and non-configured task.


fred3m (approved):
LGTM

PR #240, opened 2025-03-27
DM-49600-v29: Stand-up LATISS ingredients pipeline YAML

leeskelvin:
On DM-49600 it was discovered that the obs_lsst config overrides for the calibrateImage task in LATISS were not being picked up correctly due to a subsequent config override in `drp_pipe/config/calibrateImage.py`. Standing up a LATISS ingredients YAML and setting the appropriate config override here resolves this issue, as this file is last in the chain. The offending config has been removed from obs_lsst to prevent possible future confusion as to it being set there.


TallJimbo (approved):
Looks good, as far as one can check this kind of thing rigorously by eye.  But I don't see any other merges to `main` between the v29 branch and the commit you're trying to backport, so I'm pretty confident this is good to go.

PR #241, opened 2025-03-27
DM-49727: Replace objectEpochTable with computeObjectEpochs

taranu:


pipelines/LSSTComCam/DRP.yaml
```diff
@@ -94,6 +94,8 @@ subsets:
       - makePsfMatchedWarp
       - selectDeepCoaddVisits
       - assembleCoadd
+      - healSparsePropertyMaps
+      - computeObjectMeanEpochs
```
TallJimbo, line 98: I think the step3a tasks are supposed to be per-patch, not per-tract (and even if you take my suggestion on the pipe_tasks PR to make `computeObjectMeanEpochs` per-patch, `healSparsePropertyMaps` won't be).

I'm not sure what to do about that; it doesn't seem worth it to add a `step3c` and `step3d` just to add one more gather-scatter.  I'm hoping we can just live with it, especially because this is only about ComCam v1 and LATISS.  But we should at least make sure @yalsayyad knows and doesn't have strong feelings about it.

pipelines/_ingredients/LATISS/DRP.yaml
```diff
@@ -177,6 +177,8 @@ subsets:
       - makePsfMatchedWarp
       - selectDeepCoaddVisits
       - assembleCoadd
+      - healSparsePropertyMaps
+      - computeObjectMeanEpochs
```
TallJimbo, line 181: Same issue with these being tract-level in `step3a` as in ComCam v1.

TallJimbo (approved):

TallJimbo (commented):
If the `bands` config option sticks around I think you need to modify the `ci_imsim` and `ci_hsc` pipelines here to just use the bands in those tests.

PR #247, opened 2025-04-09
DM-49922: Fix up some clustering and add a unit test to check it

yalsayyad:


tests/test_clustering.py
```diff
@@ -0,0 +1,122 @@
...
+            # construct pipeline graph from URI
+            full_pipeline_path = os.path.join(PIPELINES_DIR, pipeline_path)
+            pipeline = Pipeline.from_uri(full_pipeline_path)
+
+            pipeline_graph = pipeline.to_graph(registry=butler.registry)
+
+            # construct bps config from clustering yaml
+            full_clustering_path = os.path.join(CLUSTERING_DIR, clustering_path)
+            with open(full_clustering_path, 'r') as f:
+                clustering_data = yaml.safe_load(f)
```
MichelleGower, line 113: I'm missing why reading from a file before passing to BpsConfig instead of letting BpsConfig read the file (i.e., `bps_config = BpsConfig(full_clustering_path)`)

tests/test_clustering.py
```diff
@@ -0,0 +1,122 @@
...
+
+        for refcat_name, dimensions, dtype, is_calibration in REFCATS:
+            self.register_refcat(butler, refcat_name)
+
+        for clustering_path, pipeline_path in CLUSTERING_PIPELINE_MAPPING.items():
+            # construct pipeline graph from URI
+            full_pipeline_path = os.path.join(PIPELINES_DIR, pipeline_path)
+            pipeline = Pipeline.from_uri(full_pipeline_path)
+
+            pipeline_graph = pipeline.to_graph(registry=butler.registry)
```
MichelleGower, line 108: I don't know if it will make a difference in future clustering tests, but I don't think a butler is required now (`pipeline.to_graph()`)

MichelleGower (approved):
Nothing stands out about the clustering changes, but I don't know the pipelines.  Very happy the small ctrl_bps changes allowed unit test to be written.  I had a couple comments about a few lines.

yalsayyad:
Thank you for pointing out those two things.  Removing the test butler reduced the lines of code in that test by half!

I also made a last minute change to wrap the call to `check_clustering_config` in a try block so that which cluster/pipeline yamls its comparing is in the error message when it fails.

PR #253, opened 2025-04-12
DM-49955:  Initial NV and DRP pipelines

yalsayyad:


pipelines/LSSTCam/nightly-validation.yaml
```diff
@@ -0,0 +1,65 @@
...
+      - photometricCatalogMatch
+      - photometricRefCatObjectTract
+      - plotPropertyMapTract
+      - validateObjectTableCore
+      - computeObjectEpochs
+      - analyzeObjectTableSurveyCore
+      - makeMetricTableObjectTableCore
+      - objectTableCoreWholeSkyPlot
+      - makeMetricTableObjectTableCoreRefCatMatch
+      - objectTableCoreRefCatMatchWholeSkyPlot
```
TallJimbo, line 46: It might be a nice pattern to declare some subsets in `base-v2.yaml` (e.g. `make-object-table`, `dia-all`) and exclude those.  Even if they aren't actually used in other places, it keeps the lists of related tasks that might need to be modified in concert together.

TallJimbo (approved):

pipelines/LSSTCam/nightly-validation.yaml
yalsayyad: That is a good idea (esp in the future when we can make subsets of subsets) but given that running `analyze-initial-visits` on its own was raising, I wasn't going to start making a subset with overlapping tasks on this ticket.

PR #254, opened 2025-04-16
DM-50211: Use AP-like configuration for calibrateImage in quickLook.

TallJimbo:


yalsayyad (approved):
Looks good. 

On DM-50000 @mfisherlevine  is using:
`config.do_calibrate_pixels = False`  I don't know whether we should keep that or not. 

and a
````  
python: |
        config.star_measurement.plugins.names.add("base_FPPosition")
```

TallJimbo:
Good points; I'll add both of those.

PR #256, opened 2025-04-19
DM-50356: Allow visit_table to be stored as astropy table but read as DataFrame

isullivan:


erykoff (approved):
While you're in here can you fix the linting error and change `False` to `false` in `pipelines/LSSTCam/quickLook.yaml`?

PR #257, opened 2025-04-20
DM-50361: Add LSSTCam quicklook pipelines for bias/dark/flat frames.

erykoff:


mfisherlevine (approved):
I wanted to change the output type, and have pushed a commit to that end. If you're OK with that then we're good to go 🙏

pipelines/LATISS/quickLookFlat.yaml
```diff
@@ -0,0 +1,19 @@
+description: |
+    A fast pipeline to be run on the rapid analysis framework at
+    the summit which feeds RubinTV for realtime observer feedback
+    specialized for LATISS for flat frames.
+instrument: lsst.obs.lsst.Latiss
+imports:
+  - location: $CP_VERIFY_DIR/pipelines/LSSTCam/verifyFlat.yaml
```
mfisherlevine, line 7: Copy & paste error - should be LATISS not LSSTCam here.

mfisherlevine: Fixed.

PR #258, opened 2025-04-21
DM-45569: Disable color fit for nightly-validation in ComCam.

PFLeget:


arunkannawadi (approved):
This is fine, but I also don't know if this is relevant anymore given that we're not even running this anymore.

PFLeget:
It was when building package it was complaining, this is why I had to add that.

PR #260, opened 2025-04-24
DM-50431: Rename finalVisitSummaryHandles

TallJimbo:
for measureObjectUnforced.


TallJimbo:
This branch seems to have been forgotten when `pipe_tasks` was merged, which broke `drp_pipe` on main.  I'm merging this now without testing because it looks like the right fix and we're in hotfix mode.

arunkannawadi:
This is indeed the right fix.

PR #262, opened 2025-04-28
DM-49008: Add hooks to switch to cell-based coadds

arunkannawadi:


pipelines/HSC/DRP-Prod.yaml
```diff
@@ -101,6 +101,7 @@ subsets:
       - makePsfMatchedWarp
       - selectDeepCoaddVisits
       - assembleCoadd
+      - assembleCellCoadd
```
arunkannawadi, line 104: Is there a way to include these only if `parameters.use_cell_coadds` is `true`?

pipelines/LSSTCam/nightly-validation.yaml
```diff
@@ -67,3 +67,7 @@ tasks:
     class: lsst.pipe.tasks.healSparseMapping.HealSparsePropertyMapTask
     config:
       connections.visit_summaries: preliminary_visit_summary
+  assembleCellCoadd:
+    class: lsst.drp.tasks.assemble_cell_coadd.AssembleCellCoaddTask
+    config:
+      connections.visitSummaryList: preliminary_visit_summary
```
arunkannawadi, line 73: Do we want to generate and run measurement on cell-based coadds for NV?

pipelines/HSC/DRP-Prod.yaml
TallJimbo: That all feels a little too magical to me, especially since I think sometimes it might be desirable to build the cell-based coadds without using them (that's what we were going to do first, right?).

If you're worried about the pain of dropping `assembleCellCoadd`  from subsets in many downstream pipelines, using `labeledSubsetModifyMode: EDIT` and `exclude: assembleCellCoadd` in the pipeline import section can probably help a lot with that (see pipelines that import `base-v2.yaml` for examples).

pipelines/LSSTCam/nightly-validation.yaml
TallJimbo: I'd lean towards doing that in a second phase, when we're ready to switch to it full in DRP; we don't want to be building two kinds of coadds in NV at any time, so we need to be ready for a complete switchover there.

pipelines/_ingredients/base-v2.yaml
```diff
@@ -265,7 +277,10 @@ tasks:
       connections.outputSchema: object_unforced_measurement_schema
       connections.outputSources: object_unforced_measurement
       connections.exposure: deep_coadd
+      connections.exposure_cells: deep_coadd_predetection_cell
```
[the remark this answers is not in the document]
TallJimbo: I'd lean towards "cells", but also no strong preference.

pipelines/_ingredients/base-v2.yaml
```diff
@@ -278,29 +293,38 @@ tasks:
...
       connections.refWcs: deep_coadd.wcs
       connections.measCat: object_forced_measurement
       idGenerator.release_id: parameters.release_id
+      useCellCoadds: parameters.use_cell_coadds
   fitDeepCoaddPsfGaussians:
     class: lsst.meas.extensions.multiprofit.pipetasks_fit.MultiProFitCoaddPsfFitTask
     config:
       connections.coadd: deep_coadd
+      connections.coadd_cell: deep_coadd_predetection_cell
+      connections.background: deep_coadd_background
```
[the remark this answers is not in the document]
TallJimbo: I think not.

pipelines/_ingredients/base-v2.yaml
```diff
@@ -278,29 +293,38 @@ tasks:
...
       connections.scarletModels: object_scarlet_models
       connections.refWcs: deep_coadd.wcs
       connections.measCat: object_forced_measurement
       idGenerator.release_id: parameters.release_id
+      useCellCoadds: parameters.use_cell_coadds
   fitDeepCoaddPsfGaussians:
     class: lsst.meas.extensions.multiprofit.pipetasks_fit.MultiProFitCoaddPsfFitTask
     config:
       connections.coadd: deep_coadd
+      connections.coadd_cell: deep_coadd_predetection_cell
```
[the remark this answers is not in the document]
TallJimbo: Either is fine with me. Someday (not at all soon), I'd like to go through a migration that would let us drop the "cell" terminology entirely, because they'd be the only kind of coadd.

pipelines/_ingredients/base-v2.yaml
```diff
@@ -989,3 +1016,6 @@ contracts:
   - reprocessVisitImage.do_apply_flat_background_ratio == calibrateImage.do_illumination_correction
   - fgcmBuildFromIsolatedStar.doApplyWcsJacobian == False if calibrateImage.do_illumination_correction else True
   - fgcmBuildFromIsolatedStar.doApplyWcsJacobian == fgcmOutputProducts.doComposeWcsJacobian
+  - (detectCoaddPeaks.useCellCoadds == deblendCoaddFootprints.useCellCoadds == measureObjectForced.useCellCoadds ==
```
TallJimbo, line 1019: Better to split this out into N-1 separate constraints, I think; that'll help narrow down what the problem is when one trips.  And it's just easier to read, and it doesn't make the user stop and think, "oh, does that actually work?"

TallJimbo (approved):

PR #265, opened 2025-05-02
DM-50000: Configure quickLook pipeline for LSSTCam on sky

mfisherlevine:


TallJimbo (approved):
Please fix the YAML lints, otherwise looks good.

PR #269, opened 2025-05-08
DM-50774: Add DRP FL pipeline

yalsayyad:
Untested DRAFT

### DIFF
--- a/pipelines/LSSTCam/DRP-FL.yaml
+++ b/pipelines/LSSTCam/DRP-FL.yaml
@@ -1,0 +1,108 @@
+description: >
+  DRP pipeline specialize for processing FL data
+
+  This pipeline is optimized for getting to deep_coadds and DIASources
+  as quickly as possible. No PIFF, no object table generation.
+  It uses SkyCorr for aesthetic coadds.
+instrument: lsst.obs.lsst.LsstCam
+imports:
+  - location: $DRP_PIPE_DIR/pipelines/_ingredients/base-v2.yaml
+    labeledSubsetModifyMode: EDIT
+    exclude:
+      - analysis-visit-initial
+      - analysis-visit-recalibrated
+      - analysis-visit-source
+      - associateIsolatedStar
+      - analyzeSingleVisitStarAssociation
+      - makeAnalysisSingleVisitStarAssociationMetricTable
+      - makeAnalysisSingleVisitStarAssociationWholeSkyPlot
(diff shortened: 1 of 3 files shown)


pipelines/LSSTCam/DRP-FL.yaml
```diff
@@ -0,0 +1,131 @@
...
+      connections.calBkgs: preliminary_visit_image_background
+      connections.calExpMosaic: preliminary_visit_image_skyCorr_visit_mosaic
+      connections.calBkgMosaic: preliminary_visit_image_background_skyCorr_visit_mosaic
+
+subsets:
+  step2c-recalibrate-visits:
+    subset:
+      - skyCorr
+    description: >
+      A step that runs visit-level per-visit background estimation
```
TallJimbo, line 126: I noticed from the pipeline graph that this could also fit into `step1c-single-visit-visits`, along with `consolidateVisitSummary` and `consolidateSingleVisitStar`.  

Reasons to leave it here, in `stage2`:
- it's conceptually very much a "recalibrate" process
- in some future pipeline I could imagine it needing to go after FGCM
- you don't have to write out the `stage1` and `step1b` subsets here just to add one task to them

Reasons to move it to `stage1`:
 - it would cluster nicely with the first set of consolidate tasks, especially `consolidateVisitSummary`, with which it could share `preliminary_visit_image` caching (sure, `consolidateVisitSummary doesn't need caching because it can do partial reads, but piggybacking on `skyCorr` caching is even better).
 - fewer steps overall

TallJimbo (approved):
I _think_ I've answered the the `reprocessVisitImage.calib_stars` connection question (`single_visit_star`), and the tests pass and the pipeline graphs build, but I haven't tried actually running it.

pipelines/LSSTCam/DRP-FL.yaml
TallJimbo: Actually I'm not so sure that shared caching would work the way we want; especially if `consolidateVisitSummary` happens to run first.  And that makes me wonder if caching for `skyCorr` is working at all.  Gonna go ask on #-dm-middleware-support.

yalsayyad: I think I'm going to leave it in stage 2 for the humans. So that we can run all of stage 1 overnight on Thursday and deal with the trying to push skyCorr through on Friday.

PR #271, opened 2025-05-10
DM-50811: Add NV-FL pipeline for interim CI

yalsayyad:
For use as a way to quickly iterate on configs
for the FL pipeline


pipelines/LSSTCam/nightly-validation-FL.yaml
```diff
@@ -0,0 +1,21 @@
+description: >
+  Temporary pipeline for validating FL data on a daily cadence
+  To be deleted after May 2025
+imports:
+  # Yes, I know we're onlu supposed to import from ingredients
```
ctslater, line 5:
```suggestion
  # Yes, I know we're only supposed to import from ingredients
```

ctslater (approved):

PR #272, opened 2025-05-10
DM-50735: run FitVisitBackgroundTask in FL pipelines

TallJimbo:
First two commits can go in whenever.  Last commit is what actually switches what `MakeDirectWarpTask` uses.

### DIFF
--- a/pipelines/LSSTCam/DRP-FL.yaml
+++ b/pipelines/LSSTCam/DRP-FL.yaml
@@ -65,7 +65,6 @@ tasks:
     config:
       doApplySkyCorr: true
       connections.visit_summary: preliminary_visit_summary
-      connections.background_revert_list: preliminary_visit_image_background
   selectDeepCoaddVisits:
     class: lsst.pipe.tasks.selectImages.BestSeeingSelectVisitsTask
     config:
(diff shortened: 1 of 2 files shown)


yalsayyad:
I missed this 16 hours ago apparently. I'll look tomorrow and we can merge tomorrow (Sunday) for Monday's run.

pipelines/LSSTCam/DRP-FL.yaml
```diff
@@ -65,7 +65,6 @@ tasks:
     config:
       doApplySkyCorr: true
       connections.visit_summary: preliminary_visit_summary
-      connections.background_revert_list: preliminary_visit_image_background
```
yalsayyad, line 68: Wait, what's going on here?

yalsayyad (approved):
Sorry, I had forgotten about this until now. So does fitVisitBackground un in ADDITION to skyCorr? I had thought it was a cheap skyCorr alternative?

Also, I want to do a skyCorr run with bgModel.xSize=2048 at some point. That example @leeskelvin showed a week ago looked good.

pipelines/LSSTCam/DRP-FL.yaml
yalsayyad: Oh nevermind, your commit message explains!

yalsayyad (approved):

PR #275, opened 2025-05-16
DM-50927: Use two nights for asteroid coadds in DRP-FL

yalsayyad:
Using the full dataset for M49 made the denominator too high to clip artifacts but not asteroids. Select visits from two very good nights of observations.

### DIFF
--- a/pipelines/LSSTCam/DRP-FL.yaml
+++ b/pipelines/LSSTCam/DRP-FL.yaml
@@ -16,7 +16,6 @@ imports:
       - analyzeSingleVisitStarAssociation
       - makeAnalysisSingleVisitStarAssociationMetricTable
       - makeAnalysisSingleVisitStarAssociationWholeSkyPlot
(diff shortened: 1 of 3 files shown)


pipelines/LSSTCam/DRP-FL.yaml
```diff
@@ -71,6 +70,15 @@ tasks:
     config:
       maxPsfFwhm: 1.6
       connections.visitSummaries: preliminary_visit_summary
+  selectTemplateCoaddVisits:
+    class: lsst.pipe.tasks.selectImages.BestSeeingQuantileSelectVisitsTask
+    config:
+      connections.visitSummaries: preliminary_visit_summary
+      nVisitsMin: 12
+      qMax: 0.8
+      # 20250502 and 20250503 were good nights
```
RobertLuptonTheGood, line 79: Please note that these are the only nights in this MJD range.

RobertLuptonTheGood (approved):
Looks good.  Seems to do what it says

PR #278, opened 2025-05-21
DM-50938: Turn on SkyCorr in LSSTCam DRP pipeline and carry over select FL configs

yalsayyad:


pipelines/LSSTCam/DRP-FL.yaml
```diff
@@ -116,26 +101,6 @@ tasks:
     class: lsst.pipe.tasks.drpAssociationPipe.DrpAssociationPipeTask
     config:
       connections.finalVisitSummaryRefs: preliminary_visit_summary
-  skyCorr:
```
isullivan, line 119: `skyCorr` can't be removed here unless you make this pipeline inherit from `DRP.yaml`, or if you add `skyCorr` to `_ingredients/base-v2.yaml`

isullivan (approved):
Note the comment on `skyCorr`, which causes the tests to fail.

I built and inspected the pipeline graph for DRP-FL as well as the stages individually, and there were no dangling datasets. I note that the outputs of `skyCorr`, `fitVisitBackground`, and `skyCorrExperimental` are not used anywhere (I don't think they are supposed to be, though).

PR #280, opened 2025-06-02
DM-51199: Add effTimeZeroPointScale visit-level cutoff

yalsayyad:


erykoff (approved):
I am thinking a bit about whether this also needs to be band dependent, but if so I guess the solution is to adjust the scaling parameters.

PR #283, opened 2025-06-18
DM-48202: LSSTComCam memory requests ad clustering yaml files

hlin730:
Add memory request and clustering yaml files


brianyanny (approved):
These changes adjust the memory requests and clustering for the LSSTComCam DP1 processing to allow jobs to go through without running out of memory.

PR #285, opened 2025-06-24
DM-50120: Turn on gbdes for LSSTCam

cmsaunders:


brianyanny (approved):
Increases default memory for gbdes from 8GB to 40GB -- should get through most/all cases without a memory overflow.
Also enables gbdesAstrometricFit now to be run as stage2b.

PR #287, opened 2025-07-03
DM-51638: LSSTCam error in analyzeSourceAssociation: "index out of bounds"

jrmullaney:
connections.associatedSourceIds in analysis-association-source.yaml was pointing to the wrong data product.


cmsaunders (approved):
LGTM

PR #293, opened 2025-07-30
DM-51971: Add DiaSource filtering task after reliability scoring

isullivan:


TallJimbo (approved):
Looks good to me.  I don't have a problem with the current names, but if you're actively looking for something shorter:
- rename the input to the new task `dia_source_unfiltered`
- rename the high-reliability output `dia_source_unstandardized`
- rename the low-reliability output `dia_source_low_reliability`

I do think what you have is about the best we can do if renaming the input to the task is not on the table.

pipelines/_ingredients/DRP-minimal-calibration.yaml
BrunoSanchez, line 1: Just asking for naming changes to remove `unstandardized` word.

BrunoSanchez (approved):

PR #295, opened 2025-08-08
DM-51830: Add config for multiprocessing with gbdes

cmsaunders:


erykoff (approved):
This looks good; but is there a concern it won't work with panda (like fgcmcal)?

cmsaunders:
According to this conversation, https://rubin-obs.slack.com/archives/C08CS8ZD83T/p1752865384680829 and the Jira ticket, DM-48156 is done, so it should work with panda. It looks like the loop still has not been closed to the extent of uncommenting out those lines in `drp_pipe`. I can do that here, or you can do it when `fgcmcal` gets turned on.

PR #296, opened 2025-08-12
DM-51687: Implement forcedPhotObject task in place of forcedPhotDirect/Diff

fred3m:


yalsayyad:
So sorry for the flyby review, but I can't tell if the "Dia" in `forcedPhotDiaDetector` is referring to the image or the objects. Meaning does it run on Difference Images or DIAObjects? 

This needs to be so beyond crystal clear. Remember that before you started this  even you and Jim were confused about which direction it was sharded when the names were: `forcedPhotDiaObjectDirect` vs `forcedPhotObjectDifference` vs `forcedPhotDiaObjectDirect` vs `forcedPhotDiaObjectDifference`.

PR #381, opened 2026-02-24
DM-54227: DRP HIPS Pipeline DRAFT

yalsayyad:


pipelines/_ingredients/analysis-hips.yaml
```diff
@@ -0,0 +1,323 @@
...
+      python: |
+        config.brightnessThresh = 0.3*16500
+  makeRGBPicturePrettyCoaddGRI:
+    class: lsst.pipe.tasks.prettyPictureMaker.PrettyPictureTask
+    config:
+      # Task default is gri
+      connections.inputCoadds: pretty_coadd_fixed_stars
+      connections.outputRGB: pretty_coadd_fixed_stars_gri
+      connections.outputRGBMask: pretty_coadd_fixed_stars_gri_mask
+  makeHighOrderHipsEpoGRI:
```
natelust: we should double check that epo only wants GRI (but maybe you already did)

natelust (commented):
I made a few comments on a per-task basis, but in general if you make changes it would apply to the various permutations of bands as well.

pipelines/_ingredients/analysis-hips.yaml
```diff
@@ -0,0 +1,324 @@
...
+      connections.inputCoadd: pretty_coadd
+      connections.outputCoadd: pretty_coadd_fixed_stars
+      python: |
+        config.brightnessThresh = 0.3*parameters.abs_max
+  makeRGBPicturePrettyCoaddGRI:
+    class: lsst.pipe.tasks.prettyPictureMaker.PrettyPictureTask
+    config:
+      # Task default is gri
+      connections.inputCoadds: pretty_coadd_fixed_stars
+      connections.outputRGB: pretty_coadd_gri
```
TallJimbo: If it's not too late to append `_array` or something to this dataset type name, that would make the transition away from a bare numpy array easier in a week or two.

(Same for all instances of this connection.)

pipelines/_ingredients/analysis-hips.yaml
```diff
@@ -0,0 +1,324 @@
...
+      connections.outputRGBMask: pretty_coadd_gri_mask
```
TallJimbo: @natelust , does this imply that I should be adding a mask plane to `lsst.images.ColorImage` over on DM-54220?

TallJimbo (approved):

pipelines/_ingredients/base-v2.yaml
```diff
@@ -1139,6 +1214,15 @@ subsets:
...
     - consolidateSsTables
+    # 4e
+    - fixStarsPrettyCoadd
+    - makeRGBPicturePrettyCoaddGRI
+    - makeHighOrderHipsEpoGRI
+    - makeLowOrderHipsEpoGRI
+    - writeAllSkyHipsInfoEpoGRI
+    - makeHighOrderHipsGRI
+    - makeLowOrderHipsGRI
+    - writeAllSkyHipsInfoGRI
```
leeskelvin, line 1242: This looks like its missing `fixBackgroundPrettyCoadd`, as per step4e.

pipelines/_ingredients/analysis-hips.yaml
```diff
@@ -0,0 +1,336 @@
+description: |
+  Default HiPS to produce during routine DRP processing.
+
+parameters:
+  base_uri: ""
+  hips_title: ""
+  abs_max: 75500
```
leeskelvin, line 7: Shouldn't this be a camera-specific override? Might be overkill, but I can see it being a potential gotcha when/if this is ever used on other cameras.

leeskelvin (approved):
A couple more comments, but again, looking fairly healthy. As in `ci_lsstcam`, don't forget to clean up the squash/fixup commits here too.

pipelines/_ingredients/analysis-hips.yaml
natelust: possibly yes, but many of our pipelines are by default tuned to this is what is run an in these ways for the lsst drp processing.  If we were to factor this out, we would need to add a whole other level of pipeline indirection because this is currently applied to all the pipelines we run things in, as this is a parameters override, not just a task config override for one camera. In principle these can be handled but in the spirit of keeping changes as minimal as possible for dp2 back porting late in the game I think it is best as is.

PR #392, opened 2026-04-16
Update Merian and DECam pipelines to include SkyCorr and ampoffset correction

zhouconghao:
The CpSky ISR ampoffset change is here: https://github.com/lsst/cp_pipe/pull/365

PR #394, opened 2026-04-21
DM-54731: Update Merian pipelines to enable running transformObjectTable

leeskelvin:


leeskelvin:
Originally defined on https://github.com/lsst/drp_pipe/pull/392.

pipelines/DECam/DRP-Merian.yaml
```diff
@@ -38,8 +38,16 @@ tasks:
...
     config:
       python: |
         config.bands=['N540', 'N708']
+  consolidateObjectTable:
+    class: lsst.pipe.tasks.postprocess.ConsolidateObjectTableTask
+    config:
+      python: |
+        config.actions.extendedness.bands = ['N540', 'N708']
+        config.actions.extendedness.bands_combined.clear()
+        config.actions.extendedness.bands_combined["N540_N708"] = "N540,N708"
```
arunkannawadi, line 53: I would strongly recommend placing the band-specific items in `obs_decam` instead of in the pipeline.

arunkannawadi (approved):

pipelines/DECam/DRP-Merian.yaml
leeskelvin: I think that might cause problems for other users of `obs_decam` who've constructed their own pipeline YAMLs. The hope here is to keep this pipeline Merian-specific.

PR #395, opened 2026-04-23
DM-54759: Fix wrong exclude name in ci_lsstcam pipeline

natelust:


laurenam (approved):
@natelust tells me this PR needs an official approval in order to to merge it, so I am hitting the button, but noting here that this is very much in the context of a "hot-fix" (i.e. we are not waiting for a Jenkins run as there isn't (awake) time to follow that through tonight and it is "Wednesday" night already...

PR #397, opened 2026-04-29
DM-53324: Update pipelines cf. new injected data names

leeskelvin:


BrunoSanchez (commented):
Lee, I see the changes, they look good to me. I agree, this is what RFC-1171 is about. 

On AP we will need to start tagging collections with fakes and making sure this is not mixed. That's all. Approved.

BrunoSanchez (approved):

PR #398, opened 2026-05-02
DM-54843: consolidateSsTables: fix G12 to 0.5, set phot. error floor to 0.01mag

mjuric:
Updates to asteroid absolute magnitude fitting for DRP pipelines:
* Fix G12 parameter to 0.5 (don't try to fit it, as we don't have much coverage)
* Mix in 0.01 mag error floor (in quadrature) as DP2 underestimates photometric errors of bright sources


isullivan (approved):
Looks good.

PR #402, opened 2026-05-16
DM-54367: Metrics of coadd masked pixel numbers at patch, tract, and sky-level

jrmullaney:


cmsaunders (approved):
This looks good to me. My only question is whether we actually want to add tasks to DRP-FL and nightly validation. 

- For DRP-FL, my impression was that the First Look pipeline should stay a snapshot of what was done for first look, but I'm not sure if that's true. If you haven't already, I would check on #lsstcam-drp-processing about whether tasks should be added there or not. 
- For nightly validation, I can see that the metrics would be useful, but I'm not sure about the plots. This seems like a case where it's going to be too many plots for people to actually be able to use them.

PR #406, opened 2026-06-04
DM-55036: Add a pipeline for converting DRP outputs to lsst.images formats.

TallJimbo:


pipelines/LSSTCam/rewrite-images.yaml
```diff
@@ -0,0 +1,50 @@
+description: >
+  A pipeline to rewrite public image data products into `lsst.images` formats,
```
timj, line 2: It's not rewriting coadds though? So do we want to hedge this description?

timj (approved):

pipelines/LSSTCam/rewrite-images.yaml
TallJimbo: Not yet, but I plan to put the coadd-rewriting into this pipeline on my current ticket.  So the description is just a little forward-looking.

PR #409, opened 2026-06-09
DM-55129: Add coadds to rewrite-images pipeline

TallJimbo:


bps/compression/LSSTCam/DRP-config_future/datastores/fileDatastore.yaml
```diff
@@ -0,0 +1,38 @@
+formatters:
+  preliminary_visit_image:
+    formatter: lsst.obs.base.formatters.fitsExposure.FitsExposureFormatter
```
timj, line 3: So preliminary_visit_image and template are always legacy but everything else has a modern version and a legacy version?

timj (approved):

bps/compression/LSSTCam/DRP-config_future/datastores/fileDatastore.yaml
TallJimbo: Yes, so far.  I'm not adding things until I have code to do the rewrite.

PR #419, opened 2026-08-14
DM-55738: Add legacy DECam configs to override new defaults

erinleighh:


mrawls (approved):
Good, this preserves existing behavior and should let existing tests pass. We can revisit changing these defaults to The Monster another time.

